PR #375, opened 2013-10-10
Ensure the message shown in the backups table when no backups have been completed spans all 4 columns

willmot:


pdewouters:
@willmot do you also get email notifs about these coveralls messages?

willmot:
Yeah, not sure where they come from...

PR #600, opened 2014-11-27
X shows when running a backup

pdewouters:
screen shot 2014-11-20 at 17 49 52

Steps to reproduce:
1. Click run backup
2. Don't reload the page
3. Wait for the backup to complete
4. Notice there is a cross instead of a tick


willmot:
@pdewouters is coveralls something you set up? Any idea why it's commenting on a 18 month old PR?

PR #657, opened 2014-12-30
Use wp_json_encode

pdewouters:
Replace occurrences of`json_encode` with `wp_json_encode`
http://pento.net/2014/12/19/json-encoding-in-wordpress-4-1/


willmot:
I'm going to close this out as it's likely going to be too stale by the time 4.3 comes around anyway.

willmot:
We can always restore the branch in the future.

Issue #246, opened 2013-03-26
backup error - open_basedir restriction in effect


venkatsrini80:
Hello,
BACKUPWORDPRESS version 2.2.1

i have installed backupwordpress plugin today of version 2.2.1 and i get below error stack while doing a backup. kindly help.

Many Thanks,
Srini

is_executable(): open_basedir restriction in effect. File(/Program Files/MySQL/MySQL Server 5.5/bin/mysqldump) is not within the allowed path(s): (C:/Inetpub/vhosts/parikshancharitabletrust.org\;C:\Windows\Temp), C:\Inetpub\vhosts\parikshancharitabletrust.org\httpdocs\wp-content\plugins\backupwordpress\hm-backup\hm-backup.php, 515

is_executable(): open_basedir restriction in effect. File(/Program Files/MySQL/MySQL Server 5.4/bin/mysqldump) is not within the allowed path(s): (C:/Inetpub/vhosts/parikshancharitabletrust.org\;C:\Windows\Temp), C:\Inetpub\vhosts\parikshancharitabletrust.org\httpdocs\wp-content\plugins\backupwordpress\hm-backup\hm-backup.php, 515

is_executable(): open_basedir restriction in effect. File(/Program Files/MySQL/MySQL Server 5.1/bin/mysqldump) is not within the allowed path(s): (C:/Inetpub/vhosts/parikshancharitabletrust.org\;C:\Windows\Temp), C:\Inetpub\vhosts\parikshancharitabletrust.org\httpdocs\wp-content\plugins\backupwordpress\hm-backup\hm-backup.php, 515

is_executable(): open_basedir restriction in effect. File(/Program Files/MySQL/MySQL Server 5.0/bin/mysqldump) is not within the allowed path(s): (C:/Inetpub/vhosts/parikshancharitabletrust.org\;C:\Windows\Temp), C:\Inetpub\vhosts\parikshancharitabletrust.org\httpdocs\wp-content\plugins\backupwordpress\hm-backup\hm-backup.php, 515

is_executable(): open_basedir restriction in effect. File(/Program Files/MySQL/MySQL Server 4.1/bin/mysqldump) is not within the allowed path(s): (C:/Inetpub/vhosts/parikshancharitabletrust.org\;C:\Windows\Temp), C:\Inetpub\vhosts\parikshancharitabletrust.org\httpdocs\wp-content\plugins\backupwordpress\hm-backup\hm-backup.php, 515

[...]

PR #652, opened 2015-01-03
Unit tests for open base dir

pdewouters:
***DO NOT MERGE***

So we can test for these kinds of errors `backup error - open_basedir restriction in effect`
Context https://github.com/humanmade/backupwordpress/issues/246

### DIFF
--- a/classes/class-hmbkp-path.php
+++ b/classes/class-hmbkp-path.php
@@ -37,44 +37,44 @@ class HMBKP_Path {
 	 */
 	protected $custom_path;
 
-    /**
-     * Protected constructor to prevent creating a new instance of the
-     * *Singleton* via the `new` operator from outside of this class.
-     */
-    protected function __construct() {}
+	/**
+	 * Protected constructor to prevent creating a new instance of the
+	 * *Singleton* via the `new` operator from outside of this class.
+	 */
+	protected function __construct() {}
 
-    /**
-     * Private clone method to prevent cloning of the instance of the
-     * *Singleton* instance.
-     *
-     * @return void
-     */
-    private function __clone() {}
+	/**
+	 * Private clone method to prevent cloning of the instance of the
+	 * *Singleton* instance.
+	 *
+	 * @return void
+	 */
+	private function __clone() {}
 
-    /**
-     * Private unserialize method to prevent unserializing of the *Singleton*
-     * instance.
-     *
-     * @return void
-     */
-    private function __wakeup() {}
+	/**
+	 * Private unserialize method to prevent unserializing of the *Singleton*
+	 * instance.
+	 *
+	 * @return void
+	 */
+	private function __wakeup() {}
 
-    /**
-     * Returns the *Singleton* instance of this class.
-     *
-     * @staticvar HMBKP_PAth $instance The *Singleton* instances of this class.
-     *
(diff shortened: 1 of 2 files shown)


pdewouters:
I have set up an `open_basedir` restriction:
screenshot 2015-01-03 20 12 48


If the folder does not exist, `file_exists()` throws a Warning:
````
[03-Jan-2015 23:13:21 UTC] PHP Warning:  file_exists(): open_basedir restriction in effect. File(/srv/backups/) is not within the allowed path(s): (/srv/backups/:/srv/www/) in /srv/www/backupwordpress.dev/content/plugins/backupwordpress/functions/interface.php on line 132
[03-Jan-2015 23:13:21 UTC] PHP Stack trace:
[03-Jan-2015 23:13:21 UTC] PHP   1. {main}() /srv/www/backupwordpress.dev/wordpress/wp-admin/admin-ajax.php:0
[03-Jan-2015 23:13:21 UTC] PHP   2. do_action() /srv/www/backupwordpress.dev/wordpress/wp-admin/admin-ajax.php:44
[03-Jan-2015 23:13:21 UTC] PHP   3. call_user_func_array:{/srv/www/backupwordpress.dev/wordpress/wp-includes/plugin.php:496}() /srv/www/backupwordpress.dev/wordpress/wp-includes/plugin.php:496
[03-Jan-2015 23:13:21 UTC] PHP   4. hmbkp_set_server_config_notices() /srv/www/backupwordpress.dev/wordpress/wp-includes/plugin.php:496
[03-Jan-2015 23:13:21 UTC] PHP   5. file_exists() /srv/www/backupwordpress.dev/content/plugins/backupwordpress/functions/interface.php:132
````

I wouldn't expect an `open_basedir` warning on a folder that is in the allowed paths, even if it does not exist.

This bug has been reported:
https://bugs.php.net/bug.php?id=53041

**Comment by willmot** [comment]
> This bug has been reported:

Nice find!

So to clarify, we get a PHP Warning if you call `file_exists` on a file / directory path that is inside allowed paths but doesn't actually exist?

Does the above also affect `is_dir` / `is_file` and it's ilk? (Wondering if we can work around).

**Comment by pdewouters** [comment]
yes it does affect those other functions too :(

**Comment by willmot** [comment]
Also, does this actually stop backups working? Or does it just cause warnings?

**Comment by pdewouters** [comment]
> Also, does this actually stop backups working? Or does it just cause warnings?

So a full backup caused a timeout error, couldn't see what caused it in the error log though, and the warnings break the admin. When the backup finishes we do a `wp_safe_redirect` which causes the "Headers already sent" error, and causes the admin to not be displayed if WP_DEBUG is enabled.

screenshot 2015-01-05 09 12 46

**Comment by pdewouters** [comment]
With `WP_DEBUG` off the manual backups run fine

**Comment by willmot** [comment]
That's good, can we just silence those errors so they don't show at all?

**Comment by pdewouters** [comment]
That's what I was thinking yeah.

File: classes/class-hmbkp-path.php [hunk 1]
```diff
@@ -37,44 +37,44 @@ class HMBKP_Path {
 	 */
 	protected $custom_path;
 
-    /**
```
pdewouters: @willmot Does your editor sometimes use spaces for indentation?

PR #475, opened 2015-01-05
Track average backup running time and pass to intercom

pdewouters:
Now that we are measuring how long backups run (see #428) for we should start tracking the average time and passing it to Intercom.


pdewouters:
screenshot 2015-01-05 15 18 52

pdewouters:
- [ ] add some unit tests

classes/class-schedule.php
```diff
@@ -965,6 +967,79 @@ protected function do_action( $action ) {
...
+	}
+
+	/**
+	 * Calculates the average run time for this schedule.
+	 *
+	 * @return string
+	 */
+	public function get_schedule_average_duration() {
+
+		if ( ! isset( $this->options['backup_run_count'] ) || ! isset( $this->options['backup_run_count'] )  ) {
```
willmot: `! isset( $this->options['backup_run_count']` is duplicated in the if statement.

classes/class-schedule.php
```diff
@@ -965,6 +967,79 @@ protected function do_action( $action ) {
...
+			return '0';
+		}
+
+		if ( 0 === (int) $this->options['backup_run_count'] ) {
+			return '0';
```
willmot: Would be good to show something other than `0` if the backup hasn't run yet as `0` makes it seem like the average run time is `0` which is probably misleading. Instead I'd show something like `N/A`.

PR #682, opened 2015-01-16
path argument interfers with wp-cli path argument.

duritong:
wp-cli also has a path argument to point to a wordpress installtion.
Hence using the path argument with backupwordpress interfers with
the global one and the wp installation won't be found anymore.
Renaming the argument to backup_path is probably the sanest option.


willmot:
Thanks for this, we actually have this fixed in the #667.

duritong:
So let's close that then?

willmot:
Given it might be awhile before we merge that branch I'm going to merge this here so it get's released in the next version.

Thanks for the contribution!

PR #683, opened 2015-01-19
User reported update error

pdewouters:
```
Parse error: syntax error, unexpected T_STATIC, expecting T_STRING or T_VARIABLE or '$' in /home4/crossq/public_html/edsung/wp-content/plugins/backupwordpress/classes/class-hmbkp-path.php on line 74
````

### DIFF
--- a/classes/class-hmbkp-path.php
+++ b/classes/class-hmbkp-path.php
@@ -37,6 +37,8 @@ class HMBKP_Path {
 	 */
 	protected $custom_path;
 
+	private static $instance;
+
 	/**
 	 * Protected constructor to prevent creating a new instance of the
 	 * *Singleton* via the `new` operator from outside of this class.
@@ -62,20 +64,17 @@ class HMBKP_Path {
 	/**
 	 * Returns the *Singleton* instance of this class.
 	 *
-	 * @staticvar HMBKP_PAth $instance The *Singleton* instances of this class.
+	 * @staticvar HMBKP_Path $instance The *Singleton* instances of this class.
 	 *
 	 * @return HMBKP_Path The *Singleton* instance.
 	 */
 	public static function get_instance() {
 
-		static $instance = null;
-
-		if ( null === $instance ) {
-			$instance = new static();
+		if ( ! ( self::$instance instanceof HMBKP_Path ) ) {
+			self::$instance = new HMBKP_Path();
 		}
 
-		return $instance;
-
+		return self::$instance;
 	}
 
 	/**

### CODE REVIEW

File: classes/class-hmbkp-path.php [hunk 1]
```diff
@@ -62,20 +64,17 @@ private function __wakeup() {}
...
 	 * Returns the *Singleton* instance of this class.
 	 *
-	 * @staticvar HMBKP_PAth $instance The *Singleton* instances of this class.
+	 * @staticvar HMBKP_Path $instance The *Singleton* instances of this class.
 	 *
 	 * @return HMBKP_Path The *Singleton* instance.
 	 */
 	public static function get_instance() {
 
-		static $instance = null;
```
pdewouters: If you initialize it to `null` here, won't it always create a new instance?

willmot: Yeah, I'd assume that too, I copied the code from here though: 

http://www.phptherightway.com/pages/Design-Patterns.html#singleton

Maybe it's "Late `static` binding"? although I don't totally have my head wrapped around that...

Maybe @joehoyle can provide insight.

joehoyle: @pdewouters `static` is a special case where it will only set it to null if it was never set before. However, I prefer the pattern of storing the static variable on the class, like so:

```PHP
class Foo {

    private static $instance;

    public static function get_instance() {
        if ( ! self::$instance ) {
            self::$instance = new HMBKP_Path();
        }
        return self::$instance;
    }
```

joehoyle: Oh looks like you used that pattern anyway

pdewouters: Yeah me too :)

willmot:
Self means it can't be overwritten by an extending class (I think) (not that that feature matters here)

pdewouters:
Isn't `self` the equivalent of `$this` in a static context? I think you are referring to the `final` keyword like in `final class MyClass`

joehoyle:
@pdewouters @willmot it does mean it can't be overriden by subclasses (as `self` refers to the class the method is on, not the called class. What you pretty much always want is `static`, however that wasn't added until PHP 5.3. All a subclass need to do override the `get_instance` method however.

pdewouters:
Yes, I understand that using `self/static` refers to class methods/properties that are available to all instantiated objects. Hadn't thought of the overriding detail though.

pdewouters:
@willmot as discussed in hangout, this can be merged for the sake of consistency and having the main plugin class not throw errors for PHP 5.2

PR #686, opened 2015-01-20
Move closing div tag

pdewouters:
The schedule sentence div was containing the schedule settings form, instead of them being on the same level

### DIFF
--- a/admin/schedule-sentence.php
+++ b/admin/schedule-sentence.php
@@ -132,7 +132,6 @@ if ( ! empty( $services ) && count( $services ) > 1 ) {
 
 	<?php require( HMBKP_PLUGIN_PATH . 'admin/schedule-settings.php' ); ?>
 
-</div>
 
 <?php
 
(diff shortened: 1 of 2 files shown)


willmot:
@pdewouters to post a screenshot so it's clear this doesn't break things visually.

pdewouters:
1. Schedule settings at same level as Schedule sentence

screen shot 2015-01-26 at 10 41 15
screen shot 2015-01-26 at 10 40 13

2. Schedule settings in the Schedule sentence div

screen shot 2015-01-26 at 10 43 17
screen shot 2015-01-26 at 10 42 33

IMO , semantically, they should be at the same level. Thoughts? Maybe you wanted them to be in the same "box" with left/right padding?

willmot:
hmm, not sure I agree.

The settings are semantically under the link to edit those settings, I think it ties them together nicely.

pdewouters:
ah, you're right - I see that now.

PR #687, opened 2015-01-21
Fix up setup class

pdewouters:
The code was not functional, this PR fixes that by also removing the need to require plugin files and use plugin methods to delete options.


willmot:
Would be great to get this unit tested

pdewouters:
@willmot these unit tests don't work. trying something else

willmot:
@owaincuvelier to do some QA to make sure activating, deactivating and uninstallation is working correctly.

You can test the cron removal by using something like [Cron Control](https://wordpress.org/plugins/wp-cron-control/)

willmot:
Looks like it needs master merging, then should be good to go.

pdewouters:
@owaincuvelier did you get a chance to review?

owaincuvelier:
@pdewouters will do this in a few minutes!

uninstall.php
```diff
@@ -1,44 +1,24 @@
...
-$schedules = HMBKP_Schedules::get_instance();
-
-// Cancel all the schedules and delete all the backups
-foreach ( $schedules->get_schedules() as $schedule ) {
-	$schedule->cancel( true );
-}
+array_map( 'delete_option', $schedules );
 
 // Remove the backups directory
+require_once( dirname( __FILE__ ) . '/../hm-backup/hm-backup.php' );
```
owaincuvelier: This no longer exists so caused fatal error

uninstall.php
```diff
@@ -1,44 +1,24 @@
...
+require_once( dirname( __FILE__ ) . '/class-hmbkp-path.php' );
```
owaincuvelier: I think the path here should be ```'/classes/class-path.php' ```

uninstall.php
```diff
@@ -1,44 +1,24 @@
...
+require_once( dirname( __FILE__ ) . '/../functions/core.php' );
```
owaincuvelier: And this path should just be ```'/functions/core.php'```. 

But still having issues with the uninstall so still testing.

uninstall.php
```diff
@@ -1,44 +1,24 @@
...
-
-// Cancel all the schedules and delete all the backups
-foreach ( $schedules->get_schedules() as $schedule ) {
-	$schedule->cancel( true );
-}
+array_map( 'delete_option', $schedules );
 
 // Remove the backups directory
+require_once( dirname( __FILE__ ) . '/class-backup.php' );
+require_once( dirname( __FILE__ ) . '/class-hmbkp-path.php' );
```
owaincuvelier: ```/class-hmbkp-path.php``` should just be ```/class-path.php```.

But this also causes a fatal error. ```Fatal error: Class 'HM\BackUpWordPress\Backup' not found in /srv/www/hmbase.dev/content/plugins/backupwordpress/classes/class-path.php on line 85```. Though I don't quite understand why it's causing a fatal error because we include the class in the line above.

This line also causes ```Use of undefined constant HMBKP_SECURE_KEY``` because the plugin is deactivated on delete, the defines presumably don't exist. But I don't think this matters?

willmot: > `require_once( dirname( __FILE__ ) . '/class-backup.php' );`

Those files are in a `classes` sub directory.

owaincuvelier: Ah right yeah. Didn't double check that. (see previous comments on old commit)

owaincuvelier: Cool, so this should be:
```
require_once( dirname( __FILE__ ) . '/classes/class-backup.php' );
require_once( dirname( __FILE__ ) . '/classes/class-path.php' );
require_once( dirname( __FILE__ ) . '/functions/core.php' );
```

owaincuvelier:
All database entires were removed except the windows azure licence entry: ```hmbkpp_waz_settings```.

All the cronjob are successfully removed on deactivation.

Just need to update the paths for ```uninstall.php``` and should be good to go.

pdewouters:
should also delete the backups directory

owaincuvelier:
>should also delete the backups directory

So we're deleting peoples backups on deletion of the plugin?

willmot:
> So we're deleting peoples backups on deletion of the plugin?

Yeah not sure we should be doing that actually

pdewouters:
Not an issue - we can add an opt-in setting "delete all data on uninstall"

PR #684, opened 2015-01-26
Look into BWP possibly deleting site htaccess file

pdewouters:
https://wordpress.org/support/topic/backupwordpress-is-deleting-wp-admin-htaccess-file?replies=2#post-6459213


pdewouters:
@owaincuvelier @willmot if you want to test the fix, you can reproduce the bug easily.  I initially thought I'd need to have a LAMP stack to reproduce but it's not necessary. Anyway if you want a local LAMP stack for WordPress I recommend Desktop Server, it's pretty neat ( don't judge by the website ) http://serverpress.com/products/desktopserver/

Steps to reproduce:

- Create a file named `.htaccess` under `wp-admin`
- Run a backup
- Notice the file is deleted
- Checkout this branch
- Run the backup
- File is still there

Not sure why this was happening though

willmot:
Would be good to understand why this was happening

pdewouters:
I should have recorded a screencast ! it's not happening anymore...

pdewouters:
Deleted previous comment, I was wrong, blame it on tiredness...

So if I var_dump `realpath($file)` I get:
  string(74) "/Users/pauldewouters/Documents/Websites/www.example.dev/wp-admin/.htaccess"

willmot:
So odd.

What about `get_path`?

pdewouters:
my guess is that `readdir` is returning a handle to the path of the calling script which would be `admin-ajax.php` that is under `wp-admin`

willmot:
Can you var dump the readdir call?

Sent while mobile

pdewouters:
````


    array(14) {

      [0] =>

      string(64) "/Users/pauldewouters/Documents/Websites/www.example.dev/wp-admin"

      [1] =>

      string(55) "/Users/pauldewouters/Documents/Websites/www.example.dev"

      [2] =>

      bool(false)

      [3] =>

      string(74) "/Users/pauldewouters/Documents/Websites/www.example.dev/wp-admin/.htaccess"

      [4] =>

      bool(false)

      [5] =>

      bool(false)

      [6] =>

      bool(false)

      [7] =>

      bool(false)

      [8] =>

      bool(false)

      [9] =>

      bool(false)

      [10] =>

      bool(false)

      [11] =>

      bool(false)

      [12] =>

      bool(false)

      [13] =>

      bool(false)

    }


````

pdewouters:
`readdir` returns `string(1) "."` which means it is scanning the current dir and not the backups dir

pdewouters:
In summary the issue is not caused by the path calculation, so we could safely merge this.

willmot:
Nice detective work!

PR #671, opened 2015-01-27
IIS: No download links

owaincuvelier:
IIS servers don't display download links.

screen shot 2015-01-08 at 14 32 29


willmot:
Instead of running `hmbkp_path` through `conform_dir` on output we should instead run the path through conform_dir inside `HMBKP_Path` as early as possible

pdewouters:
@owaincuvelier can you test on IIS please? Thanks :)

classes/class-path.php
```diff
@@ -146,13 +148,13 @@ public function get_custom_path() {
...
 
-		if ( false === $default = glob( WP_CONTENT_DIR . '/backupwordpress-*-backups', GLOB_ONLYDIR ) ) {
+		if ( false === $default = glob( WP_CONTENT_DIR . DIRECTORY_SEPARATOR . 'backupwordpress-*-backups', GLOB_ONLYDIR ) ) {
 			$default = array();
 		}
 
 		$upload_dir = wp_upload_dir();
 
-		if ( false === $fallback = glob( $upload_dir['basedir'] . '/backupwordpress-*-backups', GLOB_ONLYDIR ) ) {
+		if ( false === $fallback = glob( $upload_dir['basedir'] . DIRECTORY_SEPARATOR . 'backupwordpress-*-backups', GLOB_ONLYDIR ) ) {
```
willmot: No need for `DIRECTORY_SEPARATOR` as Windows supports `/` fine.

classes/class-path.php
```diff
@@ -82,7 +84,7 @@ public function get_path() {
 		// Ensure the backup directory is protected
 		$this->protect_path();
 
-		return Backup::conform_dir( $this->path );
```
willmot: Confused as to why you removed the `conform_dir` call here? Isn't this exactly what we want it to do?

pdewouters:
As discussed, going to give this a little more thought

willmot:
This actually seems to have been fixed by https://github.com/humanmade/backupwordpress/pull/667

pdewouters:
screen shot 2015-01-28 at 08 32 50

You're right, just installed BWP latest master on my Webmatrix install, and I can download backups

PR #726, opened 2015-02-04
Fix fatal error missing class addons

pdewouters:
BackUpWordPress 3.1.2 introduces some changes that were not backwards compatible. in particular the change of class names and introduction of namespaces. For example, the addons extend the `HMBKP_Service` class which has been renamed to `HM\BackUpWordPress\Service` This results in a fatal error.
This pull requests introduces a `deprecated/php` file which contains the old classes to avoid a fatal error.

To test, you just need older versions of the addons on your test site.


pdewouters:
@owaincuvelier if we tell users to skip 3.1.2 the addons will not throw fatal errors. Let's discuss tomorrow.

pdewouters:
Ryan suggested that the new classes extend the old ones, but there are issues with incompatible method signatures.
I think there's no harm in leaving the old classes in a separate file and throwing a deprecated warning.

willmot:
Do we actually need to keep all the old classes functional? Or is it just to prevent fatal errors?

If it's just to prevent fatals then I'd prefer to just define the minimum stuff to avoid fatals.

pdewouters:
@willmot just for Fatal errors

owaincuvelier:
PR does the job, warning looks a little ugly but its functional.

screen shot 2015-02-06 at 00 46 23

PR #727, opened 2015-02-05
Fix class name

danielbachhuber:
Reported in https://github.com/wp-cli/wp-cli/issues/1635


willmot:
Thanks @danielbachhuber, @pdewouters could you review / merge. Seems like a good candidate for next minor release

PR #732, opened 2015-02-11
Excluding any file in root will recursively exclude it from all subdirs

pdewouters:
To reproduce, you can create a `readme,txt` file under the site root then exclude it. if you inspect a backup you'll notice the readme files from plugin folders have disappeared!
Strange thing is why the unit tests haven't detected this

### DIFF
--- a/classes/class-backup.php
+++ b/classes/class-backup.php
@@ -1370,7 +1370,7 @@ namespace HM\BackUpWordPress {
 				}
 
 				// Wrap directory fragments and files in wildcards for zip
-				if ( 'zip' === $context && ( $fragment || $file ) ) {
+				if ( 'zip' === $context && ( $fragment ) ) {
 					$rule = $wildcard . $rule . $wildcard;
 				}
 
--- a/classes/class-backup.php
+++ b/classes/class-backup.php
@@ -1370,7 +1370,7 @@ namespace HM\BackUpWordPress {
 				}
 
 				// Wrap directory fragments and files in wildcards for zip
-				if ( 'zip' === $context && ( $fragment ) ) {
+				if ( 'zip' === $context && ( $fragment || $file ) ) {
 					$rule = $wildcard . $rule . $wildcard;
 				}
 
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -460,7 +460,7 @@ function hmbkp_add_exclude_rule() {
 
 	$schedule = new HM\BackUpWordPress\Scheduled_Backup( sanitize_text_field( $_GET['hmbkp_schedule_id'] ) );
 
-	$exclude_rule = str_ireplace( $schedule->backup->get_root(), '', sanitize_text_field( $_GET['hmbkp_exclude_pathname'] ) );
+	$exclude_rule = sanitize_text_field( $_GET['hmbkp_exclude_pathname'] );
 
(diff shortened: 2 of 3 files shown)


pdewouters:
@willmot @joehoyle not sure why this problem only surfaces now, but removing the wildcard prefix seems to fix it.
I'd be tempted to remove the wildcard logic as it can lead to unexpected results. Thoughts?

classes/class-backup.php
```diff
@@ -1370,7 +1370,7 @@ public function exclude_string( $context = 'zip' ) {
 				}
 
 				// Wrap directory fragments and files in wildcards for zip
-				if ( 'zip' === $context && ( $fragment || $file ) ) {
+				if ( 'zip' === $context && ( $fragment ) ) {
```
pdewouters: I'd be tempted to remove this completely as the user does not know they are excluding all folders that match the pattern. They are just excluding a specific folder by clicking the Exclude button.

willmot:
Yeah that isn't ideal.

It would be good to add in some unit tests for it (obviously they'll fail at the moment).

This is all really a hold over from when we used to support exclude rules rather than just single paths. Exclude rules are still supported through the `HMBKP_EXCLUDE` define though so I'd prefer not to lose this functionality. Ideally we'd just fix it and add some unit tests to prove it.

pdewouters:
> Exclude rules are still supported through the HMBKP_EXCLUDE define though so I'd prefer not to lose this functionality.

Yes, I know. I was including that in my thought. But I just had another idea. We could use a different approach for the constant. See for example the rules in the Gruntfile:
https://github.com/humanmade/backupwordpress/blob/master/Gruntfile.js#L108

The user would add the wildcards intentionally. Users who are already setting the constant will just have to add the wildcards after the update if they in fact did mean for it to be.

willmot:
> The user would add the wildcards intentionally. Users who are already setting the constant will just have to add the wildcards after the update if they in fact did mean for it to be.

Users can already add wildcards intentionally. The issue as I see it is that folder (IE paths / rules which end in `/`) are auto wildcarded. This needs to be the case as the rule `wp-content/` needs to also match `wp-content/uploads`. Maybe the issue is just that the above get's converted into rule `wp-content*` instead of `wp-content/*`?

I think the reason for this is that `wp-content/*` wouldn't match `wp-content` itself. We can work around that though be special casing it.

pdewouters:
To be more specific - see https://github.com/humanmade/backupwordpress/pull/732#issuecomment-73927132

To me the issue seems to be the auto prefixed wildcard which would mean any readme.txt file anywhere in the directory tree will be excluded
'*readme.txt*'

pdewouters:
@willmot let's discuss this on a quick chat tomorrow

willmot:
@pdewouters :+1:

pdewouters:
Currently, if a user sets an exclude such as `'exclude/exclude.exclude'` we make it `*'exclude/exclude.exclude*` so that it will be excluded wherever it may be in the structure. I think this assumption is wrong. We should instead consider it to be `/'exclude/exclude.exclude` as is the case with Grunt globbing patterns.

http://gruntjs.com/configuring-tasks#globbing-patterns

willmot:
Nice!

PR #735, opened 2015-02-13
README: Add information about constants

siruguri:
* Specifically, note that HMBKP_SCHEDULE_TIME overrides browser-entered
  time settings for a schedule.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -41,6 +41,12 @@ We'd also love help translating the plugin into more languages, if you can help 
 
 The plugin will try to use the `mysqldump` and `zip` commands via shell if they are available, using these will greatly improve the time it takes to back up your site.
 
+## Constants ##
+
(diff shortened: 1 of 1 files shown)


pdewouters:
@siruguri thanks for the pull request! We have detailed information about setting the constants in the help tab on the admin page BTW

siruguri:
Cool! The admin page tab doesn't explain that the constants are the final say - somehow, I feel that browser-entered data should be the last word, esp from the pov of "non techie" users. Even I had to actually read the code to confirm that in fact the `wp-config` constants were being read over the settings. Thought it might be helpful to make it explicit in the README.

pdewouters:
Good point. We should probably make it clearer to users how the contants work.
@owaincuvelier could we add a section to the support center explaining this? We could then copy it to the readme file.

owaincuvelier:
Yeah good idea.

pdewouters:
Awesome. Closing this as we have actually generate the readme file from a script.

Issue #565, opened 2014-09-17
Use settings API


pdewouters:
After 3.0 release, we could use the settings API.
We can leverage it for the settings form submission and error handling. 

Here is a good candidate:
https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-settings.php

The addons could also use it

willmot:
@pdewouters is this something that's worth doing? What are the benefits?

pdewouters:
Leveraging WordPress settings API is a benefit in itself I think. Also, would it allow us to remove some form handling code? and error handling can bedone via `settings_error` ...

willmot:
Sounds worthy of exploration 👍

PR #737, opened 2015-02-13
Use the Service name as the setting name

pdewouters:
This pull request changes the service settings name to use the slugified service name `google-drive` or `ftp` instead of `HM\BackUpWordPressGoogleDrive\`
It's just a theory that backslashes in setting names could potentially not work well in some server configs. Also it's easier to work with. For example in jQuery you have to escape the backslashes like `\\\\` 
This change requires migrating the old settings to the new settings. Also, this revealed the fact that when we added namespaces it broke the existing settings because `get_class` adds the namespace so the settings were persisted under a different option name.

### DIFF
--- a/classes/class-service.php
+++ b/classes/class-service.php
@@ -94,7 +94,7 @@ abstract class Service {
 	 * @return string       The formated name
 	 */
 	protected function get_field_name( $name ) {
-		return esc_attr( get_class( $this ) . '[' . $name . ']' );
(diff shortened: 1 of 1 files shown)


pdewouters:
Related: https://github.com/humanmade/backupwordpress/issues/565
Would require a fair amount of refactoring, but using the core settings API is a big plus for me, I think maybe a separate class for persisting service settings

pdewouters:
Closing as I'll include this in another PR with the upgrade routine

PR #739, opened 2015-02-20
Fixes settings not saving

pdewouters:
Fixes a bug with the Service settings that were not being saved because of the backslashes in the settings key names.
Also migrates any existing settings to the new naming convention. If multiple version of settings names are present in the DB, then only one will be kept.

For example, an existing site with the Google Drive addon could have 2 different serialized schedule settings for services:
`HMBKP_GDV_Backup_Service`
`HM\BackUpWordPressGDrive\Google_Drive_BackUp`

The new setting name will be `google-drive`
The upgrade routine will loop over these two and whichever one is the last to be iterated over will carry its values over to the new setting.

### DIFF
--- a/backupwordpress.php
+++ b/backupwordpress.php
@@ -3,7 +3,7 @@
 Plugin Name: BackUpWordPress
 Plugin URI: http://bwp.hmn.md/
 Description: Simple automated backups of your WordPress powered website. Once activated you'll find me under <strong>Tools &rarr; Backups</strong>. On multisite, you'll find me under the Network Settings menu.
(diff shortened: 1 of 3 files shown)


pdewouters:
@owaincuvelier when the tests have passed. would be good to test this out on the guys test site. I did a quick test by editing the file via FTP to apply the fix and it works. 
Let me know if you need more info
If we could get this and https://github.com/humanmade/backupwordpress/pull/728 merged so I can release an update

owaincuvelier:
Amazzinngg! It worked. Merged

PR #743, opened 2015-02-28
Exclude Envato-backups folders by default

pdewouters:
***Original issue***
There's also a managewp folder we can exclude. Additionally, I don't know what plugin it originated from but I saw a ```backups-*****``` folder, maybe we can catch that too.

***PR Summary***
I took a bit of a deep dive into how we were calculating directory sizes and scanning the filesystem.
I tried to consolidate our different scanning functions to use the Symfony Finder component which provides a layer of abstraction on top of the SPL Iterators.
It will allow us to do more complex "queries" in the future. It also irons out a few bugs in the default PHP Iterators and deals with cross platform inconsistencies. It's compatible with PHP 5.3

Whenever you see a directory size now, it has been calculated without the default excluded folders, for example any `.git` folder will not be taken into account for calculating the size of the folder.

### DIFF
--- a/admin/schedule-form-excludes.php
+++ b/admin/schedule-form-excludes.php
@@ -35,7 +35,7 @@
 
 						<td>
 
-							<?php if ( ( hmbkp_path() === untrailingslashit( $exclude ) ) || ( in_array( $exclude, $schedule->default_excludes() ) ) ) : ?>
+							<?php if ( ( hmbkp_path() === untrailingslashit( $exclude ) ) ) : ?>
 
 								<?php _e( 'Default rule', 'backupwordpress' ); ?>
 
--- a/admin/schedule-form-excludes.php
+++ b/admin/schedule-form-excludes.php
@@ -146,7 +146,7 @@
 
 						$root = new SplFileInfo( $schedule->backup->get_root() );
 
-						$size = $schedule->filesize( $root );
+						$size = $schedule->filesize( $root, true );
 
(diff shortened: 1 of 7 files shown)


pdewouters:
- [ ] change to default rules

screen shot 2015-02-28 at 18 12 56

pdewouters:
- [] Fix dir sizes

pdewouters:
So we have different sizes we could display for site and directories:

- For site: total size / total size minus excludes
- For each dir: total size / total size minus excludes

willmot:
> change to default rules

Can we only show those rules if those directories actually exist, otherwise might be confusing to users who have no idea what those are.

> We can show them both, e.g. (T:642MB | E:432MB) or add a Javascript toggle button

How about we just show the total amount size not counting excludes but then in a tooltip we show both with clear labelling.

`643MB in total but 352MB is excluded so only 291MB will be included in your backup`

Same for site size

`2.3GB in total but 1.8GB is excluded so only 535MB will be included in your backup`

Thoughts?

pdewouters:
screen shot 2015-03-03 at 15 32 22

As we don't actually scan to see if they exist and they are wildcard patterns, I just hid them.
I think this is good enough for this PR.
I'd rather focus on other issues, we can always come back and fine tune this later

willmot:
Great idea!

pdewouters:
\0/ :shipit:

PR #698, opened 2015-03-02
Check mysql user actually has access to the database via mysqldump

pdewouters:
_From @willmot on January 21, 2014 19:56_

Sometimes the mysql user specified in `wp-config.php` might not actually have access to the database via `mysqldump`.

````
mysqldump: Got error: 1045: Access denied for user \'jci_jciuser\'@\'localhost\' (using password: YES) when trying to connect
````

_Copied from original issue: humanmade/hm-backup#57_

### CODE REVIEW

**Comment by pdewouters** [comment]
So, on line 538 we check if mysqdump works, but if it's in an alternate path, we don't check it , so I added the same check there. 
I also added some more recent version of mySql for Windows

**Comment by willmot** [comment]
This ticket is about checking whether the mysql user has access to the database via the command line though by trying to actually connect to the DB

**Comment by pdewouters** [comment]
Could it be that the server is configured to disallow any kind of CLI commands? Not specific to user?

**Comment by pdewouters** [comment]
> mysqldump requires at least the SELECT privilege for dumped tables, SHOW VIEW for dumped views, TRIGGER for dumped triggers, and LOCK TABLES if the --single-transaction option is not used. Certain options might require other privileges as noted in the option descriptions.

**Comment by pdewouters** [comment]
@owaincuvelier for now - I'd tell the user to check the mysql privileges for their user via phpMyAdmin.

**Comment by pdewouters** [comment]
Did a little digging:
There are 2 important options here: `lock-tables` and `single-transaction`
Both require the `LOCK TABLES` privilege otherwise a permission denied is thrown.
However this will throw a different error: `Got error: 1044: Access denied for user 'testuser'@'localhost' to database 'backupwordpress' when using LOCK TABLES`

It might raise an issue with mixed schemas as `single-transaction` will only lock tables with `myinnodb` format. So we should check that when a user reports an errorwhen they try restoring a DB.

To get back to the issue at hand, I can reproduce  the error when I use an incorrect password. So I wonder, when can this occur? How can the DB password be incorrect but the user can still log into WordPress? He should get a "database inaccessible" error page

**Comment by pdewouters** [comment]
Do we want to display an admin notice?

**Comment by willmot** [comment]
> Do we want to display an admin notice?

I don't think so as the backup will just fallback to a PHP database dump.

File: classes/class-backup.php [hunk 1]
```diff
@@ -1,6 +1,7 @@
 <?php
 
 namespace HM\BackUpWordPress {
+	use Symfony\Component\Process\Process;
```
willmot: Does `Process` need adding requiring in `composer.json`?

pdewouters:
- [ ] fallback to PHP

classes/class-backup.php
willmot: Ignore, seen this below.

pdewouters:
review merge

classes/class-backup.php
```diff
@@ -1828,6 +1839,37 @@ public function error_handler( $type ) {
...
+
+			$cmd = "mysql --host={$host} --user={$user} --password={$pwd} {$db}";
+
+			// Pipe STDERR to STDOUT
+			$cmd .= ' 2>&1';
+
+			// Store any returned data in an error
+			$stderr = shell_exec( $cmd );
+
+			if ( $stderr ) {
```
willmot: I think we will always get an error on MySQL > 5.6 because we are passing password via CLI (like here: https://github.com/humanmade/backupwordpress/pull/698/files#diff-0c135420eb3271449b66b83a8aa89035R880

PR #748, opened 2015-03-03
Fix incorrect type error

pdewouters:
This fixes a bug where we were trying to use `conform_dir` on an SPL object instead of on a path string.


willmot:
👍

PR #734, opened 2015-03-04
Multisite backup backs up the site as many times as there are sites

pdewouters:
So after further investigation, @willmot's assumption was correct. We were adding schedules for each site on the network instead of only for the main site.
This fixes the issue by checking if we are on the main site before scheduling.

- [x] Clear non main site schedules on upgrade. Otherwise users will have to use a plugin to do it manually ( WP Crontrol )


pdewouters:
Simple fix, only tricky part is getting rid of the duplicate schedules on upgrade

functions/core.php
```diff
@@ -230,6 +230,38 @@ function hmbkp_update() {
...
 
 		array_map( 'delete_transient', $transients );
+
+		// Clear duplicate schedules on multisite
+		if ( is_multisite() ) {
+
+			global $wpdb;
+
+			// get current blogs from DB
+			$blogs = $wpdb->get_results( "SELECT blog_id FROM {$wpdb->blogs} ORDER BY blog_id" );
```
willmot: any reason  we can't use `get_sites` here?

pdewouters: do you mean `wp_get_sites()`? other than the fact I just found out about it? :)

willmot: 👍

functions/core.php
```diff
@@ -230,6 +230,38 @@ function hmbkp_update() {
...
+				switch_to_blog( get_current_blog_id() );
+
+				if ( is_main_site( get_current_blog_id() ) ) {
+					continue;
+				}
+
+				global $wpdb;
+
+				// Get the schedule options
+				$schedules = $wpdb->get_col( $wpdb->prepare( "SELECT option_name FROM $wpdb->options WHERE option_name LIKE %s", 'hmbkp_schedule_%' ) );
```
willmot: Thinking we should also delete all `hmbkp_%` options on each of those sites?

Maybe we could just run `uninstall` on all but the main site?

pdewouters: good call

pdewouters: With `uninstall.php` can I call a plugin funciton? If so, I could add an uninstall function that we can call so we don't repeat the code.
Also, `delete_option` and `delete_transient` work on the curren blog right?

willmot: You can, but you need to include the file that contains the function directly as the plugin isn't loaded when `uninstall.php` is run.

pdewouters: probably not worth the effort

willmot: 👍

PR #750, opened 2015-03-06
Schedule time format is inconsistent

pdewouters:
We should adda description and placeholders telling the user that they need to use 24H format, and display the time as such in the sentence.
screen shot 2015-03-04 at 17 40 52


pdewouters:
screen shot 2015-03-06 at 15 14 02

Users have no choice but to enter the hours in 24  hour format as we don't provide a AM/PM selection. However we can display the next schedule time in the format from Settings General.
Just added a note about this below the hour/minute inputs.

PR #759, opened 2015-03-10
Refactor main plugin file to avoid errors for PHP 5.2

pdewouters:
We should just move the whole class to an included file and only keep a PHP version check in the main files to avoid all these support requests from people on PHP 5.2


willmot:
👍

PR #757, opened 2015-03-10
Large backups finish but schedule-running not deleted

pdewouters:
I ran a 3GB backup on my vagrant box, and it finishes ok, but the status file is never deleted so it looks like the backup is running. should be an easy fix

This PR improves how we handle manual backups started from the admin. Previously we made a GET request via Ajax and waited for the server response. This could cause timeout errors.
Now we kick off a Backdrop task and return immediately. Any errors that occur during backup are stored in a DB option and when the schedule finishes, the page will reload and display any errors.
Also changed the order for complete backups, we now zip the DB dump first because with large backups, the process of including the dump in the archive a posteriori fails.

screen shot 2015-03-11 at 11 53 48


pdewouters:
- [ ] Need to get the schedule errors in a format compatible with the Notices class.

willmot:
> We wait for the response so that we can capture any Fatal Error output.

As we're going to lose this, let's make sure we open a ticket to bring it back, probably with `register_shutdown_function`

admin/actions.php
```diff
@@ -72,60 +72,34 @@ function hmbkp_request_do_backup() {
 		check_admin_referer( 'hmbkp_run_schedule', 'hmbkp_run_schedule_nonce' );
 	}
 
-	// Fixes an issue on servers which only allow a single session per client
-	session_write_close();
```
willmot: I think we'll want to keep this as Backdrop still counts as a session right?

admin/actions.php
```diff
@@ -72,60 +72,34 @@ function hmbkp_request_do_backup() {
...
+	HM\BackUpWordPress\Path::get_instance()->cleanup();
 
-	// We want to display any fatal errors in this ajax request so we can catch them on the other side.
-	error_reporting( E_ERROR | E_USER_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_RECOVERABLE_ERROR );
```
willmot: We'll lose access to these errors, we should open a new ticket to add this functionality back in, probably via `register_shutdown_function`

willmot:
This looks good 👍

admin/actions.php
pdewouters: maybe, Backdrop calls `wp_remote_post` on the admin-ajax file. 
I added it back

willmot:
Nice work @pdewouters 👍

PR #479, opened 2015-03-12
No need to choose hour on hourly backup

pdewouters:
When setting up an hourly schedule it only actually makes sense to choose the minutes as the backup will simply run every hour.

screen shot 2014-03-14 at 12 13 37


pdewouters:
So I just hid both fields when user selects hourly backup. I think that should suffice

PR #751, opened 2015-03-18
Change plugn name to include backup as a keyword

pdewouters:
Plugins are ranked by exact search term match. So if you search for backup or backups, the first results will be plugins with "backup or "backups" as a word in them
Not sure if we need to rebrand or just change the name:

Not sure if we can use `BackUp WordPress`, so maybe  `BackUpWordPress Backup Plugin`, other plugin authors are already using the trick.

@willmot @noeltock @owaincuvelier thoughts?

### DIFF
--- a/backupwordpress.php
+++ b/backupwordpress.php
@@ -1,6 +1,6 @@
 <?php
 /*
-Plugin Name: BackUpWordPress
+Plugin Name: BackUpWordPress Backup Plugin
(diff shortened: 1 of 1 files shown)


willmot:
Choo choo 🚄

pdewouters:
all aboard the fail train :)

willmot:
That's the merge train!

pdewouters:
phew

PR #770, opened 2015-03-19
backup missing cache.php file from wp-inclides directory

pdewouters:
user reported this. maybe because we ex


willmot:
Nice catch!

PR #773, opened 2015-03-20
Fixes the zip method backup

pdewouters:
The `user_can_connect` method was wrongly coded. It established a connection but then didn't quit so it made it impossible to execute any other commands.
This PR fixes that by adding the `--execute` option with a value of `quit` to quit the mysql command prompt immediately.

Also it makes the `FS` option a user defined constant because it was causing too many issues.
It also adds some user defined constants for forcing the zip method. Useful for debugging.

### DIFF
--- a/classes/class-backup.php
+++ b/classes/class-backup.php
@@ -953,17 +953,17 @@ namespace HM\BackUpWordPress {
 		public function archive() {
 
 			// Do we have the path to the zip command
-			if ( $this->get_zip_command_path() ) {
+			if ( ( defined( 'HMBKP_FORCE_ZIP_METHOD' ) && ( 'zip' === HMBKP_FORCE_ZIP_METHOD ) ) || $this->get_zip_command_path() ) {
 				$this->zip();
 			}
 
 			// If not or if the shell zip failed then use ZipArchive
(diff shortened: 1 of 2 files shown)


pdewouters:
@owaincuvelier once the Travis build passes you could test this locally and on a user site if we have access to one.

willmot:
We'll ask want to stop copying the previous backup unless HMBKP_SYNC IS defined.

pdewouters:
> We'll ask want to stop copying the previous backup unless HMBKP_SYNC IS defined.

I can't seem to find that code

willmot:
Oh actually, it doesn't duplicate the old backup, it just passes it in here: https://github.com/humanmade/backupwordpress/blob/master/classes/class-schedule.php#L783

I think we are fine to leave that code as is though as I think we'll want to bring `-FS` back once we can properly detect whether it's supported or not.

In-fact, instead of a define, why not just disable entirely for now. Or maybe it's easy enough to test if `-S` is supported?

pdewouters:
you'd have to actually run a backup to test if it works I think

owaincuvelier:
@pdewouters what do I need to do to reproduce the bug locally and test the fix?

PR #780, opened 2015-03-28
Added colons and several periods

michaelbeil:
Assigned to @pdewouters.

I am also wondering if it might be benefical to mention what branch the contributor should PR against. There is no dev, so master seems to be the obvious choice.


pdewouters:
Good idea. Yes PRs are made against master

michaelbeil:
K gonna add that.

PR #786, opened 2015-04-01
Fixes full backup tests

pdewouters:
When I refactored the plugin in 3.2.0 , I made it so `get_files` would just skip default excludes completely. However, `get_files` is used in `get_included_files` which is used by the full backup unit tests functions to test if the resulting zip archive contains the same number of folders and files as there actually should be.
As the full backup tests don't run anymore on Travis - they are completely excluded, I didn't notice that some default exclude patterns were too greedy. This led to some core files being excluded. Including some BackUpWordPress files ( `backups.php` and `backup-table.php` )

Should we revert to running full backup tests on Travis?


pdewouters:
not sure why the bump version commit is here, it's already on master

willmot:
> Should we revert to running full backup tests on Travis?

The main reason I removed them was that they took ages to run. Happy to re-enable them and work on speeding them up instead.

PR #788, opened 2015-04-02
Get home path should return defined value

mattheu:
I'm trying to run this on a multisite with WP in a subdirectory. This is only possible if I can define the HMBKP_ROOT - as this is incorrectly calculated otherwise.

However `get_home_path` doesn't use the defined value. If this is added here - it all works fine.


pdewouters:
Thanks @mattheu , much appreciated! I ran the full backup tests locally and they pass. So merging.

PR #797, opened 2015-04-15
[WIP] Wraps the setting and getting of the transient in json_encode

pdewouters:
Core has switched to JSON encoding. It’s faster and it’s more human-readable in the database
Also avoids some security issues relating to serialised objects, although that’s not relevant here

- [ ] probably need to delete the existing transient on activation

### DIFF
--- a/classes/class-schedule.php
+++ b/classes/class-schedule.php
@@ -456,7 +456,7 @@ class Scheduled_Backup {
 	public function recursive_filesize_scanner() {
 
 		// Use the cached array directory sizes if available
-		$directory_sizes = get_transient( 'hmbkp_directory_filesizes' );
+		$directory_sizes = json_decode( get_transient( 'hmbkp_directory_filesizes' ) );
 
 		// If we do have it in cache then let's use it and also clear the lock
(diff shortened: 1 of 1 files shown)


willmot:
> Core has switched to JSON encoding.

Could you link up the background on this in terms of Core (trac ticket or Make post etc)

pdewouters:
No, @johnbillion mentioned it in a Slack convo, maybe he has more insight

johnbillion:
There's just a general move to using JSON encoding instead of relying on PHP object serialisation which has security and performance concerns. https://core.trac.wordpress.org/ticket/25311 is one example off the top of my head.

pdewouters:
has something changed since @sanchothefat opened this ticket https://core.trac.wordpress.org/ticket/23753 , it seems related.

johnbillion:
The difference there is that WordPress can't switch to natively using JSON encoding in `[add|update|get]_option()` because it's lossy if someone stores a PHP object in there. Switching to selectively using JSON encoding where appropriate is preferable though.

pdewouters:
Not clear on benefits of this.

PR #798, opened 2015-04-15
Fixes activation and deactivation hooks not being fired

pdewouters:


willmot:
doh, just realised this was in-progress, sorry!

pdewouters:
I was going to add some tests.

I should have prefixed the subject with WIP but also usually I only assign
it to you and tag it with review merge when it is ready

PR #799, opened 2015-04-16
Fix activation

pdewouters:
Not sure the tests are correct because if I comment out the calls to `register_activation_hook` and `register_deactivation_hook`, the tests still pass. Although if I comment out the line `deactivate_plugin` in the deactivation test, it fails...

@johnbillion @willmot thoughts?

### DIFF
--- a/tests/other/testUninstallActivateDeactivate.php
+++ b/tests/other/testUninstallActivateDeactivate.php
@@ -10,18 +10,14 @@ class testUninstallActivateDeactivateTestCase extends HM_Backup_UnitTestCase {
 
 
 	public function setUp() {
-
+		parent::setUp();
 	}
 
 	public function tearDown() {
-
+		parent::tearDown();
 	}
 
-	public function testDeactivation() {
-		
-	}
-
-	public function testUninstall() {
+	public function test_uninstall() {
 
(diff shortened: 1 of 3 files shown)


tests/other/testUninstallActivateDeactivate.php
```diff
@@ -38,4 +34,24 @@ public function testUninstall() {
...
 
 	}
+
+	public function test_deactivate() {
+
+		$res = activate_plugin( 'backupwordpress/backupwordpress.php' );
+		$this->assertFalse( is_wp_error( $res ) );
+		deactivate_plugins( 'backupwordpress/backupwordpress.php' );
+		$this->assertFalse( is_plugin_active( 'backupwordpress/backupwordpress.php' ) );
+		$this->assertGreaterThanOrEqual( 1, did_action( 'deactivate_backupwordpress/backupwordpress.php' ) );
```
johnbillion: You could test that this equals `10` instead, as that's what it should be.

pdewouters: `did_action` returns the number of times the action fired.

johnbillion: Ah my mistake. I was thinking of `has_action()` which returns the priority of the callback.

PR #806, opened 2015-04-25
[WIP] Deactivates the plugin if it has been activated

pdewouters:
Users have been reporting that after upgrading, their admin has been inaccessible due to the PHP version check.
However I have tried to reproduce locally by installing and activating version 3.0 on a PHP 5.2 install and then upgrading to the latest version and the plugin deactivates itself. So I'm not able to reproduce.


pdewouters:
Maybe wontfix this one. Doesn't seem to be an issue anymore

PR #809, opened 2015-04-27
Error if variable is wrong type

pdewouters:
`Warning: array_key_exists() [function.array-key-exists]: The first argument should be either a string or an integer in /home/content/52/9722152/html/wp-content/plugins/backupwordpress/classes/class-schedule.php on line 416`


pdewouters:
@willmot this should fix it by returning `0` on error, it will be handled by the `elseif`

willmot:
Looks good 👍

Would be good to have this unit tested!

PR #813, opened 2015-04-29
Fixes admin page slug for multisite

pdewouters:
I'm not sure why, but the multisite `get_current_screen` returns `settings_backupwordpress-network` not `settings_backupwordpress`


pdewouters:
@owaincuvelier should work on multisite now.

PR #822, opened 2015-06-05
[WIP] Correctly set the `sslverify` argument

pdewouters:
See https://github.com/wp-cli/wp-cli/pull/1677 for explanation

### DIFF
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -685,6 +685,16 @@ function hmbkp_ajax_cron_test() {
 
 	}
 
+	// Skip test if Cron is disabled
+	if ( defined( 'DISABLE_WP_CRON' ) && DISABLE_WP_CRON ) {
+
+		delete_option( 'hmbkp_wp_cron_test_failed' );
+
+		echo 1;
+
+		die;
+	}
+
 	// Skip the test if they are using Alternate Cron
 	if ( defined( 'ALTERNATE_WP_CRON' ) ) {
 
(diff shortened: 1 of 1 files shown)


willmot:
I'm not sure that this makes sense.

We always want to set `sslverify` to `false`.

pdewouters:
why?

willmot:
Because it causes the cron check to incorrectly fail on servers which have a broken ssl setup.

pdewouters:
I don't get it, sorry.

pdewouters:
Seems the check doesn't ignore broken SSL setups anyway: https://wordpress.org/support/topic/ignore-self-signed-certificates?replies=6#post-7049321

willmot:
Ok, sounds like we should follow what https://github.com/humanmade/Backdrop/pull/10 and https://github.com/wp-cli/wp-cli/pull/1677 are doing, happy to go with that 👍

This PR also does a bunch of other things though which I'm not sure about.

- It no longer runs multiple times to avoid false positives.
- It switches to `wp_remote_post` from `wp_remote_head`.

What was the reasoning behing those changes?

pdewouters:
- [ ] revert back to multiple tests and use `wp_remote_head`

pdewouters:
@willmot don't you think 30 seconds would be too long a timeout setting?

willmot:
> @willmot don't you think 30 seconds would be too long a timeout setting?

Too long for what?

The goal of the timeout is simply to ensure the request has enough time to complete even on slow hosts.

I don't think there's much of a downside to a longer timeout as this task just runs in the background and reports the warning to the user when it's done.

15 would likely be ok, maybe 30 seconds is a little long.

admin/actions.php
```diff
@@ -696,22 +706,28 @@ function hmbkp_ajax_cron_test() {
 
 	}
 
-	$url = site_url( 'wp-cron.php' );
+	$spawn1 = hmbkp_get_cron_spawn();
+	$spawn2 = hmbkp_get_cron_spawn();
+	$spawn3 = hmbkp_get_cron_spawn();
+
+	$code1    = wp_remote_retrieve_response_code( $spawn1 );
+	$message1 = wp_remote_retrieve_response_message( $spawn1 );
```
willmot: Seems these `$message` variables aren't used?

pdewouters: Ah yes, 
screen shot 2015-07-18 at 14 40 10

willmot:
@pdewouters there are a bunch of other changes in this PR that fall outside the specific change mentioned in the title and I'm unsure of the rational behind them.

For example, this PR could have achieved it's goal by just changing:

```PHP
$url = site_url( 'wp-cron.php' );

// Attempt to load wp-cron.php 3 times, if we get the same error each time then inform the user.
$response1 = wp_remote_head( $url, array( 'timeout' => 30 ) );
$response2 = wp_remote_head( $url, array( 'timeout' => 30 ) );
$response3 = wp_remote_head( $url, array( 'timeout' => 30 ) );
```

To

```PHP
$url = site_url( 'wp-cron.php' );
$args = array (
	'timeout'  => 15,
	'sslverify' => apply_filters( 'https_local_ssl_verify', version_compare( $wp_version, 4.0, '<' ) ),
);

// Attempt to load wp-cron.php 3 times, if we get the same error each time then inform the user.
$response1 = wp_remote_head( $url, $args );
$response2 = wp_remote_head( $url, $args );
$response3 = wp_remote_head( $url, $args );
```

Are the other changes important from a functional point of view? If so, in what way? If not, are they mostly stylistic / re-factoring? How can we be sure they don't introduce issues?

Overall I'm not a fan of changing things unless we are clear on why and how they improve the plugin and ideally have tests to prove it. (Although I absolutely recognise our tests are a work in progress, but until they are more comprehensive I think it would be more prudent to avoid unnecessary changes).

admin/actions.php
```diff
@@ -685,6 +685,16 @@ function hmbkp_ajax_cron_test() {
...
+
+		delete_option( 'hmbkp_wp_cron_test_failed' );
+
+		echo 1;
+
+		die;
+	}
+
 	// Skip the test if they are using Alternate Cron
 	if ( defined( 'ALTERNATE_WP_CRON' ) ) {
```
willmot: We should also check if `ALTERNATE_WP_CRON` is `true` like you're doing for `DISABLE_WP_CRON`

admin/actions.php
```diff
@@ -696,22 +706,28 @@ function hmbkp_ajax_cron_test() {
 
 	}
 
-	$url = site_url( 'wp-cron.php' );
+	$spawn1 = hmbkp_get_cron_spawn();
+	$spawn2 = hmbkp_get_cron_spawn();
+	$spawn3 = hmbkp_get_cron_spawn();
+
+	$code1    = wp_remote_retrieve_response_code( $spawn1 );
```
willmot: Why switch to calling `wp_remote_retrieve_response_code ` here rather than in the `array_map` on https://github.com/humanmade/backupwordpress/pull/822/files#diff-b2d8a71a75e0cf40115b3529a31e0761L712

admin/actions.php
```diff
@@ -728,3 +744,37 @@ function hmbkp_ajax_cron_test() {
...
+
+/**
+ * Spawn a request to `wp-cron.php` and return the response.
+ *
+ * This function is designed to mimic the functionality in `spawn_cron()` with the addition of returning
+ * the result of the `wp_remote_post()` request.
+ *
+ * @return WP_Error|array The response or WP_Error on failure.
+ */
+function hmbkp_get_cron_spawn() {
```
willmot: I'm not clear why we need this function instead of just the `wp_remote_head` requests we were doing inline at https://github.com/humanmade/backupwordpress/pull/822/files#diff-b2d8a71a75e0cf40115b3529a31e0761L702 seems like it introduces a lot of new logic / filtering options that could impact how this works?

admin/actions.php
```diff
@@ -728,3 +744,37 @@ function hmbkp_ajax_cron_test() {
 
 }
 add_action( 'wp_ajax_hmbkp_cron_test', 'hmbkp_ajax_cron_test' );
+
+/**
+ * Spawn a request to `wp-cron.php` and return the response.
+ *
+ * This function is designed to mimic the functionality in `spawn_cron()` with the addition of returning
+ * the result of the `wp_remote_post()` request.
```
willmot: Needs updating to reference `wp_remote_head` now that we've changed that in the function.

pdewouters:
Closing this as not the right approach. No need to introduce the `hmbkp_get_cron_spawn` function.

Would it be better to call the core function `spawn_cron()` instead of doing a `wp_remote_head` ?

pdewouters:
- [ ] delete branch

PR #828, opened 2015-07-02
Use namespaced functions

skithund:
Fixes two errors when trying to run `wp backupwordpress backup`

```
PHP Fatal error:  Class 'Backup' not found in wp-content/plugins/backupwordpress/classes/class-backupwordpress-wp-cli-command.php on line 57
PHP Fatal error:  Class 'Path' not found in wp-content/plugins/backupwordpress/classes/class-backupwordpress-wp-cli-command.php on line 63
```


pdewouters:
Thanks

PR #830, opened 2015-07-13
Prevent duplicate entries in cron schedule option lists + language updates

ucavus:
I can see the `hmbkp_` prefix was used to separate this plugin's cron schedules from others'. However, this pollutes the global option list with duplicates, affecting the UI of other plugins. My changes retain an internal schedule list but remove these duplicates.

Also added some general language corrections and additional English files.


pdewouters:
Thanks for your pull request!

I see what you mean about the duplicates. It would be great to fix this, although I'll need to check why we have separate schedules for existing default WordPress ones in the first place.

However, there are lots of unrelated changes in this diff which makes it hard to review.
Could you split it into smaller PRs, for example:

- PR only with cron schedule name fix
- PR only with string/typo fixes
- PR with refreshed lang files

Not change function names if unnecessary
Make sure build tests pass.

Thanks

Example of issue
screen shot 2015-07-13 at 07 47 38

ucavus:
Not easily, don't I have to do separate brnaches so that new commits aren't automaticlly added to the pull request?

Has it failed or just not passed, *yet*?

pdewouters:
I think we'll settle for just renaming the labels, something like `BWP Once Daily` as we need to keep them separate from the default intervals.

ucavus:
They *are* still accessed separately. All the functions that called the plugin-specific list only get the original list. If you're worried about extra items appearing in it, drop the `apply_filter()` line I added. The changes are backwards compatible, because the upgrade drops the prefixes.

Being content to add duplicate items to a core list seems like a bit of a cop-out. At least try out the changes in your own git repo by pulling directly from me into a temporary branch before dismissing them.

pdewouters:
Thanks, I'll give this a go locally soon

ucavus:
Cool. Another thing I realised is that, if someone selects an `hmbkp` prefixed cron schedule (because there's no discernible difference between duplicates to the user), uninstalling your plugin will make their schedule fail, rather than fall back to the existing value.

pdewouters:
> if someone selects an hmbkp prefixed cron schedule

Selects it where?

There would need to be an upgrade routine for the jobs scheduled with the prefixed interval to be moved to the new interval.

willmot:
Just want to pop in and list the goals for BackUpWordPress here:

- We don't want to use the default schedule list directly as other plugins can mess with it thus messing things up for users of BWP
- We don't want the custom BackUpWordPress schedules to impact core or other plugins.

Seems we are failing on that second goal at the moment.

Can we use cron withou using the standard WordPress schedule list at all? Or does using wp-cron require that we add our custom schedules to that core list?

ucavus:
@pdewouters Selects a prefixed schedule for any other plugin that lists cron schedules. Must have missed that one, can fix it.

@willmot I would think not. Otherwise, how would the core wp-cron know how to map the schedule key to the interval if it's not added during the `cron_schedules` filter.

classes/class-schedule.php
```diff
@@ -690,17 +688,7 @@ public function get_interval() {
...
-
-		// remove any schedule whose key is not prefixed with 'hmbkp_'
-		foreach ( $schedules as $key => $arr ) {
-			if ( ! preg_match( '/^hmbkp_/', $key ) ) {
-				unset( $schedules[ $key ] );
-			}
-		}
-
-		return $schedules;
+		return hmbkp_cron_schedules();
```
willmot: Might it make sense to just move the contents of the `hmbkp_cron_schedules` function into `::get_cron_schedules` and do away with `hmbkp_cron_schedules` altogether? It's `static` so can easily be used externally.

functions/core.php
```diff
@@ -271,6 +251,26 @@ function hmbkp_update() {
 		}
 	}
 
+	// Update from PRIOR_VERSION
+	if ( get_option( 'hmbkp_plugin_version' ) && version_compare( NEXT_VERSION, get_option( 'hmbkp_plugin_version' ), '>' ) ) {
+
+		$schedules = HM\BackUpWordPress\Schedules::get_instance();
+
+		// Loop through all schedules and re-set the reccurrence to include hmbkp_
```
willmot: This comment is now incorrect

functions/core.php
```diff
@@ -271,6 +251,26 @@ function hmbkp_update() {
 		}
 	}
 
+	// Update from PRIOR_VERSION
+	if ( get_option( 'hmbkp_plugin_version' ) && version_compare( NEXT_VERSION, get_option( 'hmbkp_plugin_version' ), '>' ) ) {
```
willmot: `NEXT_VERSION` isn't defined anywhere, is that a placeholder?

pdewouters: @willmot yes, it is specified somewhere that this needs to be defined

functions/core.php
```diff
@@ -336,14 +336,18 @@ function hmbkp_setup_default_schedules() {
...
+		'monthly'     => array( 'interval' => 30 * DAY_IN_SECONDS, 'display' => __( 'Once Monthly', 'backupwordpress' ) ),
+	);
 
-	$schedules['hmbkp_hourly']      = array( 'interval' => HOUR_IN_SECONDS, 'display' => __( 'Once Hourly', 'backupwordpress' ) );
-	$schedules['hmbkp_twicedaily']  = array( 'interval' => 12 * HOUR_IN_SECONDS, 'display' => __( 'Twice Daily', 'backupwordpress' ) );
-	$schedules['hmbkp_daily']       = array( 'interval' => DAY_IN_SECONDS, 'display' => __( 'Once Daily', 'backupwordpress' ) );
-	$schedules['hmbkp_weekly']      = array( 'interval' => WEEK_IN_SECONDS, 'display' => __( 'Once Weekly', 'backupwordpress' ) );
-	$schedules['hmbkp_fortnightly'] = array( 'interval' => 2 * WEEK_IN_SECONDS, 'display' => __( 'Once Biweekly', 'backupwordpress' ) );
-	$schedules['hmbkp_monthly']     = array( 'interval' => 30 * DAY_IN_SECONDS, 'display' => __( 'Once Monthly', 'backupwordpress' ) );
+	apply_filters( 'hmbkp_cron_schedules', $schedules );
```
willmot: I'm don't think there's any need for this filter as people can simply use the standard `cron_schedules` filter to add their own?

languages/backupwordpress-en_AU.po
```diff
@@ -0,0 +1,1211 @@
+# Copyright (C) 2015 Human Made Limited
```
willmot: I'm not sure what purpose this file serves, it's a blank translation to Australian English? If you were trying to update the original English strings then you need to update the `.pot` file rather than adding a `.po` and `.mo`

languages/backupwordpress-en_GB.po
```diff
@@ -0,0 +1,1211 @@
+# Copyright (C) 2015 Human Made Limited
```
willmot: Again, not sure what the point of haivng this is? Unless I'm missing something

willmot:
This looks solid, thanks for the conribution @ucavus! There's a few things to resolve above and then happy to merge.

functions/core.php
willmot: 👍

functions/core.php
ucavus: I suggested dropping this in order to retain your own set list.

languages/backupwordpress-en_AU.po
ucavus: Not blank, just few lines translated (biweekly is ambiguous and not widely used outside the US). I thought that WP uses the binary, which contains the translated lines sans empties.

willmot:
Thanks @ucavus, we'll be discussing this in our [weekly dev chat](https://github.com/humanmade/backupwordpress/blob/master/CONTRIBUTING.md#weekly-dev-chat) on Friday and then merging 👍

willmot:
Discussed in [Slack](https://backupwordpress.slack.com/archives/general/p1437746047000159)

Agreed to:

- [ ] Implement the suggestions from the inline code comments above
- [ ] Move the copy changes to a separate issue, they'll have translation file impacts as we'll need to regenerate the original strings.
- [ ] Drop the translation files that translate `biweekly` to `fortnightly`, we should just default to `every two weeks`
- [ ] Correct version in the upgrade routine

pdewouters:
Moved dupe fix to https://github.com/humanmade/backupwordpress/pull/863

pdewouters:
Closing in favor of separate PRs mentioned aboce

PR #831, opened 2015-07-13
This should speed up imports when there's no mysqldump available (alter table disable keys etc)

palepurple:
When creating a dump file, disabling the keys on the table will speed up import by a reasonable amount. This patch might support this....

### DIFF
--- a/classes/class-backup.php
+++ b/classes/class-backup.php
@@ -1549,7 +1549,7 @@ namespace HM\BackUpWordPress {
 			$sql_file .= "#\n";
 			$sql_file .= "\n";
 			$sql_file .= "DROP TABLE IF EXISTS " . $this->sql_backquote( $table ) . ";\n";
-			$sql_file .= "ALTER TABLE " . $this->sql_backquote( $table ) . " DISABLE KEYS \n";
+			$sql_file .= "ALTER TABLE " . $this->sql_backquote( $table ) . " DISABLE KEYS; \n";
 
 			/* Table Structure */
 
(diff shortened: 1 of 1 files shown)


pdewouters:
Thanks for the Pull Request
Do you mean it should speed up exports? We never import anything.
Do you have any stats to back this up?

palepurple:
Hi - Yes ... I meant that after applying this patch, when you go to import a database dump created by this plugin it should be quicker...

Before :  ~25 minutes.
After: about 19 seconds...

willmot:
We're nervous to change the PHP mysqldump code as it currently has 0 automated tests.

Is there anything to be concerned about in terms of `DISABLE_KEYS` only being supporting in specific versions of MySQL?

We're also exploring moving to a better tested PHP mysqldump implementation: https://github.com/humanmade/backupwordpress/pull/835

willmot:
@palepurple it actually looks like we're going to switch PHP mysqldump implementations, see https://github.com/humanmade/backupwordpress/pull/835.

It would be really helpful if you could test out that branch, maybe this PR ends up no longer being relevant?

palepurple:
@willmot  - hopefully I'll have time tomorrow to give it a whirl, thanks for the pointer.

palepurple:
I've tried the mysqldump-php branch, and the .zip seems to be the same as what my PR produces (alter keys, extended insert etc).  However, it doesn't seem to list backups in the web ui that do exist on the FS. Otherwise all good.

pdewouters:
- [ ] Verify this https://github.com/humanmade/backupwordpress/pull/831#issuecomment-122817148
- [ ] test on dreamhost / pantheon

willmot:
I think we can close this in favour of #835

Issue #819, opened 2015-05-14
Add support for dismissable notices


pdewouters:
Use new dismissable notices
https://make.wordpress.org/core/2015/04/23/spinners-and-dismissible-admin-notices-in-4-2/

willmot:
Is this referring to the new Core support for dismiss buttons? If so, I agree, let's move to Future Release

willmot:
We now have these.

Issue #538, opened 2014-09-01
Correctly handle directories that are actually custom stream wrappers


pdewouters:
For example, if the site is using the S3 uploads plugin to store media on Amazon S3, BWP complains:

```
BackUpWordPress is almost ready. The backups directory can't be created because your wp-content directory isn't writable, run chown www-data:www-data s3:/hmn-uploads/newton/uploads or chmod 777 s3:/hmn-uploads/newton/uploads or create the folder yourself.
```

> s3:/hmn-uploads/newton/uploads is writable, but I think probaly fails the PHP function we are using

willmot:
cc @joehoyle would be good to get your input here

joehoyle:
After looking into the whole BUWP + S3 Uploads thing, the main problem is that BUWP is not going to support an uploads directory using a custom stream handler (that's to say anything other than the filesystem stream handler.)

I'd say you should have an activated check or whatever that the uploads dir is a local stream wrapper, and fail if it is not. This error is just the beginning of what the problems will currently be even if suppressing this error.

willmot:
Makes sense, I'm +1 on detecting and warning the user. Does the fact the the uploads folder is using a custom stream wrapper always mean we need to fail though or will it only fail if we are trying to store backups in the uploads directory?

If the backup is stored elsewhere on the filesystem it should work fine right? In that case, are we able to backup the uploads directory if it's using a custom stream wrapper? Or does it stop us accessing the files for the purposes of backup also?

I think we have two possible scenarios both of which need detecting and reporting to the user.

1. Your backups directory couldn't be created in `WP_CONTENT_DIR` and thus tried to fallback to creating it in uploads but couldn't because it's using a custom stream wrapper. Because of this we can't perform backups until you manually define a valid backups directory. This is basically the same as the generic, we couldn't create you a backups directory error message we already have.
2. Your backups are stored elsewhere so that's fine, but we're unable to backup your uploads folder (or whatever folder) because it's using a stream wrapper.

Make sense?

joehoyle:
Right, so if your using a custom stream wrapper, I guess there's no reason you can't store the backups in `/tmp` or whatever, however if you do that, I don't know how backupwordpress supporting downloading those backups.

However, if you're not backing up the uploads, I'm not sure of the utility for anything other than database backups. 

[...]

PR #833, opened 2015-07-16
Remove folder permissions notices

pdewouters:
The newly added Path class is smart enough to create the backups folder either under `wp-content` or under the `uploads` folder which will always be writable by WordPress.

The `whoami` command is sometimes disabled on shared hosts. This can cause unnecessary errors for users.

I don't think we should use `WP_Filesystem` to create folders / files as the backup functionality is mostly run without user interaction, so displaying a login form doesn't help.

`wp_mkdir_p` should suffice, we should probably throw an error and bubble it up to UI if it is unsuccessful though.

https://github.com/humanmade/backupwordpress/blob/master/classes/class-path.php#L211


pdewouters:
@willmot any thoughts?

willmot:
I agree we should remove those error messages, the advice to run `chmod` and `chown` is completely useless to most users and bad practise anyway.

However we'll still have an issue if the backups directory can't be created (which is the only time users see those above messages anyway.

I would see us using `WP_Filesystem` as part of some kind of setup screen or maybe from an action shown in the error message, IE it's a one time thing which the user initiates to use `WP_Filesystem` to create the backups folder with the correct permissions.

willmot:
Discussed this in Slack:

We should:

- Try to create the folder in `WP_CONTENT` using `wp_mkdir_p`
- If not possible, attempt to use `WP_Filesystem` to create it (with user entering their details)
- If not possible (or we need a fallback and can't ask user for details) then fallback to uploads

pdewouters:
Fixes #855 
Fixes #819

pdewouters:
@willmot do you  think this is a good approach?

willmot:
@pdewouters will need to check it out locally, could you drop in some screens of the user flow.

pdewouters:
screenshot 2015-09-04 10 48 53

To replicate, you'll need to `chmod 000 /backups-folder` but you can't do that from inside vagrant.
It will then display the FTP creds form until it gets correct credentials, then it will create the folder.

pdewouters:
- [ ] hide everything else on page and explanation message

willmot:
More like:

screen shot 2015-09-25 at 15 02 57

See https://github.com/humanmade/backupwordpress/pull/872 for the page heading "BackUpWordPress"

willmot:
@pdewouters tested this locally by removing write permissions from my `content` and `uploads` directories

```
sudo chmod 554 content
sudo chmod 554 content/uploads
```

Couple of issues:

screen shot 2015-10-05 at 10 03 03

- [ ] The form is shown twice, should only show once.
- [ ] The form is shown above the page header and error message, it should be below.
- [ ] The error message could be clearer.
- [ ] I have a second error message `ERROR: There was an error connecting to the server, Please verify the settings are correct.`. I hadn't submitted the form, this just shows when I visit the page.

pdewouters:
Ah, I must have tested it incorrectly then, thanks

pdewouters:
I'm reworking this a bit

willmot:
@pdewouters I'd suggest we bump to 3.4

pdewouters:
Ok, I'll just cherry pick the notices fixes.

willmot:
Just noting, here's what we get on hmn.md:

screen shot 2016-01-20 at 12 50 31

This is both this issue and https://github.com/humanmade/backupwordpress/issues/538

willmot:
@pdewouters I picked this back up last night and started brining back to life 👍

pdewouters:
ah cool

willmot:
Made some good progress, still a couple of things left to do:

- [ ] Attempt to create in uploads if we can't create in `WP_CONTENT_DIR`.
- [ ] Good error message for if it's still fails to create a folder.

This PR simply attempts to create the `mkdir`, it doesn't first try to set permissions on the parent directory. Setting permissions does sometimes help, but runs the risk of causing issues if the permissions  were like there were for good reason. For example `WP_CONTENT_DIR` could be own by a different user with `0775` permissions, if we then corrected them to `0755` we'd actually be making the permissions more restrictive.

- [ ] Explore setting permissions to permissive, creating directory and then setting back.

willmot:
> Explore setting permissions to permissive, creating directory and then setting back.

Decided not to go this route for now as there's no good reason that the permissions wouldn't allow the FTP user to write files.

willmot:
@pdewouters I'm ready for a review merge here.

I tested this locally by issuing the following commands from inside Vagrant:

```
sudo chmod 554 wp-content
sudo chmod 554 wp-content/uploads
```

I think you need to be using the vmware-vagrant-providor though as virtual box doesn't support modifying file permissions.

functions/interface.php
```diff
@@ -166,25 +169,21 @@ function set_server_config_notices() {
...
-				$messages[] = sprintf( __( 'Your custom backups directory %1$s isn\'t writable, new backups will be saved to %2$s instead.', 'backupwordpress' ), '<code>' . esc_html( HMBKP_PATH ) . '</code>', '<code>' . esc_html( Path::get_path() ) . '</code>' );
+				$messages[] = sprintf( __( 'Your custom backups directory (%1$s) isn\'t writable, new backups will be saved to %2$s instead.', 'backupwordpress' ), '<code>' . esc_html( HMBKP_PATH ) . '</code>', '<code>' . esc_html( Path::get_path() ) . '</code>' );
 
 			}
 		}
 	}
 
 	if ( ! is_readable( Path::get_root() ) ) {
-		$messages[] = sprintf( __( 'Your site root path %s isn\'t readable.', 'backupwordpress' ), '<code>' . Path::get_root() . '</code>' );
+		$messages[] = sprintf( __( 'Your sites root path (%s) isn\'t readable. Please contact support.', 'backupwordpress' ), '<code>' . Path::get_root() . '</code>' );
```
pdewouters: `site\'s`

willmot: Thanks, fixed in 3267c87

pdewouters:
@willmot yeah, couldn't manage to change permissions and don't really want to switch to vmware. I thought I managed to do it through the mac finder before... happy to merge though

willmot:
I'm going to try an test on a server that doesn't have direct write access I think 👍

willmot:
Here's the flow:

1. Install plugin normally, navigate to backups page.
2. If the backup directory can't be created you now see this:
screen shot 2016-03-18 at 22 36 19
3. If you enter incorrect server details you see an error
screen shot 2016-03-18 at 22 38 40
4. If you enter the correct server details but we still fail to create the backups directory you see this error:
screen shot 2016-03-18 at 22 36 28
5. If the backups could be created the page reloads and you see the normal backups admin.

@katmoody in addition to show the message and ftp form we also link people to this page for help on how to manually create their backups path: https://bwp.hmn.md/support-center/backupwordpress-faqs/#where it's pretty out of date though, as part of this ticket being merged we should get that updated.

willmot:
@dashaluna fancy merging master in here and also testing this locally (you'll need Salty WP setup to use vmware instead of virtualbox though).

dashaluna:
@willmot started looking into this:

1. Merged master
2. Cleanup code a bit - mainly indentation and simplifying `if/elseif` statements.

*TODO*:
- [ ] Setup environment for testing
- [ ] Test

functions/interface.php
```diff
@@ -107,7 +106,7 @@ function admin_notices() {
 
 				<?php foreach ( $notices['server_config'] as $notice ) : ?>
 
-					<li><?php echo wp_kses_data( $notice ); ?></li>
+					<li><?php echo print_whitelist_html( $notice ); ?></li>
```
dashaluna: @rmccue so I've got a question here. Below in this file, I'm storing translated strings in vars, each has has it's own let of allowed HTML and some don't have allowed HTML. So far so good. However, what do I use here as a generic escaping function to display text with varied set of allowed HTML. Does it make sense?

To adhere to `escape late` WP rule, here must be some escape function here. But then, how to I cater for strings with varied set of allowed HTML? In your [Whitelist HTML docs](https://github.com/humanmade/whitelist-html) there is example how to save sage HTML to a variable, but there is no example how to correctly echo it afterwards. It would be great if you could add it to the docs :)

Thanks!

rmccue: It sounds like you want to output essentially arbitrary HTML here, in which case, there's not a really good option. Passing `'comment'` as the third param will give you just tags allowed in comments, which is generally safe. However, it doesn't allow anything not in that list, obviously.

Late escaping is about trying to avoid outputting unsafe code, but sometimes you want the ability to output arbitrary HTML. It seems like in this case, you want to do that, so just `echo` would probably be fine.

If you've already made sure the variable is safe before it gets here but phpcs is throwing an error, you can "open the escape hatch" and add a `// XSS OK` comment just before the `?>`. This tells phpcs that you're sure there's no XSS possibilities here and you're safe. Be very careful doing this though. :)

dashaluna: Yes, ideally I don't wanna do that - because it because harder to track and make sure that all the strings are properly escaped - grrr :( I wonder if I can save strings together with an HTML list of allowed elements and then extract that data just before running `print_whitelist_html()` is that a better approach?

willmot: I think it would be fine to just whitelist some common HTML tags here. `strong`, `b`, `i`, `em`, `code` would probably cover everything for now. Can always add others as needed.

dashaluna: and `a`? There are lots of cases where we refer either to BWP docs or PHP docs. However, in some cases we want to allow URL to be localised and in some now. How do we tackle that?

willmot: Yeah a too

dashaluna: Ok, I'll give it a go :)

dashaluna:
@katmoody re @willmot's comment here https://github.com/humanmade/backupwordpress/pull/833#issuecomment-198572538 Just wanted to check if the instructions have been updated?

dashaluna:
*Tested*

1. All default permission => created a backup folder in `wp-content`.
2. `wp-content` set to 554 => created a backup folder in `wp-content/uploads`.
3. `wp-content` and `wp-content/uploads` both set to 554 => Error message, see screenshot.
screen shot 2016-07-27 at 12 19 42
4. Connection information provided incorrectly => Error message, see screenshot.
screen shot 2016-07-27 at 12 45 11
5. Connection information provided correctly. Tested SSH2. => Error message.
screen shot 2016-07-27 at 12 56 56
Then, I clicked on `Tools -> Backups` I was presented with the connection form again.
**NB**: @willmot how do I test other connection types?
6. Added custom folder `bwp-backups` to the root and to `wp-config.php`, tried both settings:
`define( 'HMBKP_PATH', 'bwp-backups' );`
`define( 'HMBKP_PATH', '/srv/www/wordpress-develop.dev/src/bwp-backups' );`
Default permissions are `drwxr-xr-x`

Still presented with `Connection Information` screen. @willmot  what I'm doing wrong here? The docs are not really helpful for this part.

*Logs errors*

1. Loads of these ones => PHP Warning:  fileperms(): stat failed for /srv/www/wordpress-develop.dev/src/wp-content/backupwordpress-5a644d272f-backups in /srv/www/wordpress-develop.dev/src/wp-content/plugins/backupwordpress/classes/class-requirement.php on line 282

*Docs notes* /cc @willmot @katmoody

Re https://bwp.hmn.md/support-center/backupwordpress-faqs/#where This is a bit unclear. Things to improve:

- [ ] improve php snippet formatting from `define (‘HMBKP_PATH’ , ’your directory’);` to `define( 'HMBKP_PATH', 'your_directory' );` to adhere to WP coding standards.
- [ ] link directly to section where it's explained how to change `wp-config.php`.
- [ ] explain where a backup dir should be created manually, at the root (?) if permissions are denied for `wp-content` and `wp-content/uploads`?

dashaluna:
While at it, I decided to submit a trac ticket, so that the `Content Information` form is designed in line with other Admin forms, as it looks very sad at the moment. Trac ref: https://core.trac.wordpress.org/ticket/37484

willmot:
> Connection information provided correctly. Tested SSH2. => Error message. 

This should now work correctly, since https://github.com/humanmade/backupwordpress/pull/833/commits/3f812480ca230b100164adefa022244a55e81273

> NB: @willmot how do I test other connection types?

It's not hugely important to as `WP_Filesystem` should provide consistency across them all. But, to test FTP, you'll need to setup an FTP server on your Vagrant running as the same user as `php-fpm`.

> Added custom folder `bwp-backups` to the root and to `wp-config.php`, tried both settings:

Fixed, custom backup paths are now handled separately. (aside, you need to specify the absolute path to the directory you want, and it needs to be writable by `php-fpm`.

> **Logs errors**

Fixed.

> **Docs notes**

I guess this is superseded by the new knowledge base, @katmoody?

willmot:
@dashaluna if you could retest 6. in your testing above to just double-check that it's now working as expected that'd be great. Then I think this is good to go!

dashaluna:
**Retesting previous steps** NB: gonna copy over the text for clarity, but not screenshot. Scroll up for previous test for them.

1. `Pass` All default permission => created a backup folder in wp-content.

2. `Pass` wp-content set to 554 => created a backup folder in wp-content/uploads.

3. `Pass` wp-content and wp-content/uploads both set to 554 => BWP Error message as expected.

4. `Pass` Connection information provided incorrectly => Error message as expected.

5. `Pass` Connection information provided correctly. Tested SSH2. =>
Results:
- Redirected to the main BWP admin page.
- `wp-content` permissions were fixed to `drwxr-xr-x`.
- a backup folder was created in `wp-content` folder.

`Error` I've decided to make a back up with `Run now` => the spinner is running and then replaced with a tick, however, there is no backup in the WP Admin list nor any files in the backup folder on the server. Tried with files and DB separately. Don't see any errors anywhere.

@willmot does `Run now` means do the backup now despite the schedule?

6. `Error` Added custom folder `bwp-backups` to the root dir and added the following settings to `wp-config.php`:

`define( 'HMBKP_PATH', '/srv/www/wordpress-develop.dev/src/bwp-backups' );`
Default permissions are `drwxr-xr-x`

Going to the main BWP admin page => Error message, see screenshot.
@willmot I don't know how I can check and make the newly created directory writable by `php-fpm`. Any tips? 
screen shot 2016-09-01 at 14 00 57

I've added a dirty hack to check if the directory is writable.. `echo is_writable( '/srv/www/wordpress-develop.dev/src/bwp-backups' );` and it displayed `1` which means that it is. Which in turn means that either I've tested incorrectly or it doesn't work.

willmot:
> @willmot does Run now means do the backup now despite the schedule?

It does, that should work. Could you test running a backup on master and perhaps with this branch but without changing backup path permission just to see if it's a regression caused by the work in this PR or a separate issue.

> 6. **`Error`** Added custom folder `bwp-backups` to the root dir and added the following settings to `wp-config.php`:

From the screenshot it doesn't look like the define has been picked up, are you definitely adding it to the correct `wp-config.php` file?

katmoody:
@dashaluna and @willmot - Here is where I initially updated the current FAQs when I separated them out for the new Knowledge Base section - but I was not sure it satisfied everything Dasha noted above: https://bwp.hmn.md/knowledge-base/where-are-my-backups-stored/

willmot:
@katmoody that help text looks right, although the last part of the second paragraph reads a bit odd.

>  Using the "HMBKP_PATH" constant you can change the directory and even where to store your backups by adding to the wp-config.php file using the “HMBKP_PATH” variable. 

- `HMBKP_PATH` is a constant not a variable.
- It's odd to say "you can change the directory and even where to store the backups as they are both one and the same.

I'd simplify to something like:

> You can change the directory that your backups are stored in by defining the `HMBKP_PATH` constant in your `wp-config.php` file like so:

dashaluna:
`Run now` tested on:

1. **`Pass`** `master` branch on `VirtualBox` server
2. **`Pass`** `remove-whoami` (current) branch on `VirtualBox` server
3. **`Error`** `master` branch on `VMWareFusion` server, default backup folder. The spinner is appearing and I see the tick, but no backups in the WP admin or on the server.
4. **`Error`** `remove-whoami` (current) branch on `VMWareFusion` server, default backup folder. The spinner is appearing and I see the tick, but no backups in the WP admin or on the server.

@willmot is it something wrong with my `VMWareFusion` server setup?

dashaluna:
> > Error Added custom folder bwp-backups to the root dir and added the following settings to wp-config.php:

> From the screenshot it doesn't look like the define has been picked up, are you definitely adding it to the correct wp-config.php file?

@willmot I'm adding it to the main `wp-config.php` file. I've just tested it with Virtual Box as well - the same story, the specified `HMBKP_PATH` isn't being picked up for me. Here is the screenshot:
screen shot 2016-09-21 at 12 53 08

dashaluna:
> @dashaluna and @willmot - Here is where I initially updated the current FAQs when I separated them out for the new Knowledge Base section - but I was not sure it satisfied everything Dasha noted above:https://bwp.hmn.md/knowledge-base/where-are-my-backups-stored/

@katmoody could we also change the code example to the following, to be in line with WP coding standards:

`define( 'HMBKP_PATH', '/absolute/server/path/to/your/backup/folder' );`

willmot:
> @willmot I'm adding it to the main wp-config.php file.

Can you confirm that `wp-config.php` is definitely the one being used (just try deleting it's contents or something). Maybe there's a `wp-config.php` in `src`.

dashaluna:
> > @willmot I'm adding it to the main wp-config.php file.

> Can you confirm that wp-config.php is definitely the one being used (just try deleting it's contents or something). Maybe there's a wp-config.php in src.

*Face palm* it is indeed the `src/wp-config.php` file, I didn't notice it was there..

dashaluna:
@willmot you were right, I had to edit `src/wp-config.php` file. I've tested the `Run now` functionality and updated this comment https://github.com/humanmade/backupwordpress/pull/833#issuecomment-248574296 with the test results.

It doesn't seem to do it on `VMWareFusion` server. I've no idea why :( Can we ask someone else to test this on `VMWareFusion`? Or should we look at my instance of it and figure out why it's not saving backups?

willmot:
I'm running Fusion locally and can confirm it's working fine. Do the unit tests run when using vmware for you?

willmot:
@dashaluna I went through and retested the whole flow here and everything is working as expected. Shall we screenshare to debug your Fusion issues?

willmot:
@sambulance @shadyvb this one just needs conflict resolving and can then be merged.

sambulance:
Part of the merge conflict was because we removed the `is_backup_possible()` check from around the admin recently (https://github.com/humanmade/backupwordpress/issues/1030). I've created a segmented version of this to just check if we have access to read and write directories, and if not, show the credentials form.

sambulance:
@shadyvb Please can you review/merge this, specifically my last commit?

katmoody:
@sambulance is this completed?

sambulance:
This is waiting on code review, doesn't have to be Shady.

katmoody:
I think this is still looking for a review per Sam's last response.

willmot:
Yep this looks good 👍

PR #835, opened 2015-07-17
[WIP] Adds PHP mysqldump fallback lib

pdewouters:
Now points to `dev-master` to get latest commits


willmot:
This is a large change to a really important part of complex, core functionality, we need to be confident that merging this will:

1. ~~Works~~ Unit tests pass which is a solid indication that it does.
2. Improves our PHP fallback database backups
3. Cause as few unintended breakages as possible

@pdewouters how much testing etc. has gone into this so far? What else do you think we need to do before we decide to merge and release?

pdewouters:
At present, I ran a DB only backup and tested the import.

willmot:
Couple of quick thoughts

- [ ] I notice it requires PDO, from some quick Googling it seems PDO is a default extension from PHP 5.1+, is having PDO as a dependancy going to be an issue?
- [ ] Given this class aims to mirror `mysqldump` we should make an effort to use the same settings for it as we use for `mysqldump`.

I'm a pretty big +1 on this, I think it will give us a much more robust PHP mysqldump implementation with a lot more consistency between it and the real `mysqldump`, it should also fix a bunch of bugs (both known and unknown) with our current implementation.

But we should do more testing first.

pdewouters:
> is having PDO as a dependancy going to be an issue
Would you rather leave the current function in for servers that don't support PDO?

willmot:
I guess first thing is we should try to work out whether PDO is ever not available, if it's always there then no need for a fallback, I'd strongly prefer not to have to leave the current implementation in as well.

willmot:
- [ ] Another thought, do we need to specifically pass in `DB_CHARSET` to support databases that don't use the default (`utf8`)

pdewouters:
looks like you can pass quite a few params https://github.com/ifsnop/mysqldump-php#dump-settings

pdewouters:
author has tagged a new release https://github.com/ifsnop/mysqldump-php/releases/tag/v1.6

willmot:
Let's talk through our merge plan for this in our Friday dev chat, I want to be sure we've covered our bases so that we don't cause a spike in support issues post release.

willmot:
@markoheijnen I know you've done a bunch of looking into PDO etc. Is it safe to assume all PHP 5.1 > installs have PDO support? Or should we be providing a fallback for people that don't?

willmot:
In terms of moving towards merge, the open questions are

**Do we need to provide a fallback if `PDO_MYSQL` isn't enabled?**
 - [ ]@pdewouters to test on some common hosts
 - [ ] Start tracking PDO support in Intercom https://github.com/humanmade/backupwordpress/issues/848
 - [ ] Reach out to @markoheijnen

markoheijnen:
There are some changes made in PHP in the beginning and I believe the PDO library could be disabled in 5.1 unsure about 5.2. I do know that people had issues with it but I never experienced that. What people do say is that since PHP 5.3 and later should be more then fine.

Note that I played with it on a lower scale then this plugin will do.

willmot:
Cools cheers @markoheijnen, we require 5.3 so I think we'll be fine.

pdewouters:
bedankt ;)

pdewouters:
Developers can change the dump options by filtering `hmbkp_mysqldump_fallback_dump_settings`, for example if you wanted to switch the `single-transaction` flag to false:
````
add_filter( 'hmbkp_mysqldump_fallback_dump_settings', function( $options ) {
  return array_merge(
    $options,
    array( 'single-transaction' => true ),
  )
});
````

willmot:
Nice, do you think we need to separate filters for `mysqldump` and `mysqldump_fallback` or can they just use the same settings now?

pdewouters:
We could probably abstract out the configuration into its own function and use it in both cases, yes

pdewouters:
So, for socket connections, we'll still need to parse the `DB_HOST` value and just construct a DSN string from it to pass as a parameter to the constructor. Not sure when this will be implemented though.
I guess we'll have to set this on hold?

ifsnop:
Hi there,

In the new version of mysqldump-php (take a look at the devel branch), ia now possible to connect using sockets, using the standard dsn string.

Here is the function that parses the dsn string and extracts dbname, host, socket, etc:

https://github.com/ifsnop/mysqldump-php/blob/devel/src/Ifsnop/Mysqldump/Mysqldump.php#L213

Hope that it helps.

pdewouters:
cheers @ifsnop , really appreciate that you added the feature so fast :+1: 
We have our weekly dev chat later and will discuss how to best test it.

ifsnop:
Think in using mysqldump-php not only as a fallback, but as a complete
solution, and forget calling the mysqldump binary.   You will get a php
only solution,  easier to maintain.

willmot:
I've been wondering the same, my only concern would be performance, especially when dumping large DB's (say > 1GB), have you done any benchmarking to measure performance?

pdewouters:
yeah , I was thinking exactly the same!

ifsnop:
Could anyone provide a testing dataset so we could all share benchmarcks?

I've been wondering the same, my only concern would be performance,
especially when dumping large DB's (say > 1GB), have you done any
benchmarking to measure performance?

—
Reply to this email directly or view it on GitHub
<https://github.com/humanmade/backupwordpress/pull/835#issuecomment-126638661>
.

pdewouters:
https://github.com/humanmade/backupwordpress/issues/856

ifsnop:
I did some  testing:
~~~
CREATE TABLE `test000` (
  `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
  `guid` varchar(36) NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  `email` varchar(255) DEFAULT NULL,
  `category` varchar(50) DEFAULT NULL,
  `text` text,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
~~~
Filled with some random values:
~~~
+------------+--------+------------+------------+----------------+---------------+---------------+------------+
| table_name | engine | row_format | table_rows | avg_row_length | total_mb      | data_mb       | index_mb   |
+------------+--------+------------+------------+----------------+---------------+---------------+------------+
| test000    | InnoDB | Compact    |     336192 |           5046 | 1618.00000000 | 1618.00000000 | 0.00000000 |
+------------+--------+------------+------------+----------------+---------------+---------------+------------+
~~~
Small example:

~~~
INSERT INTO `test000` VALUES (1,'6608D7AC-E472-AB1C-3A4C-8F173B8ADCA2','Leigh','in.hendrerit.consectetuer@diamatpretium.org','New
South Wales','Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Curabitur sed tortor. Integer aliquam adipiscing lacus. Ut nec urna et arcu imperdiet ullamcorper. Duis at lacus. Quisque purus sapien, gravida non, sollicitudin a, malesuada id, erat. Eti am vestibulum massa rutrum magna. Cras convallis convallis dolor. Quisque tincidunt pede ac urna. Ut tincidunt vehicula risus. Nul.... 
~~~

Results:
~~~
# time mysqldump test003 > mysqldump_test003.sql
real    0m13.578s
user    0m10.236s
sys     0m1.388s

# time php test.php
real    0m32.219s
user    0m22.788s
sys     0m4.104s
~~~

Please do you own tests!

pdewouters:
Another argument for this is that the current function uses `mysql_` function instead of `mysqli_` or PDO functions.

pdewouters:
I can't get my hands on a really large database for testing this, any suggestions?

willmot:
We're going to merge this, @pdewouters just doing final work to support Socket connections and then it's going in \0/

pdewouters:
@willmot updated the library to `2.0` , it now requires a PDO connection string. 
I tested a socket connection locally, `define('DB_HOST', 'localhost:/var/run/mysqld/mysqld.sock');`
you can easily find the path to the socket in the phpinfo output

willmot:
@pdewouters great, ready to ship?

pdewouters:
@willmot can you remind me about this?

> Make sure we are passing in the same args as we pass to mysqldump.

ifsnop:
Recently a bug was discovered, and databases with uppercase letters weren't
parsed. No release was made, but it has already been commited to master.

In a few hours I'll make a release, adding some new dump options.

Regards,

On Thu, Oct 22, 2015 at 12:48 PM, Paul de Wouters <notifications@github.com>
wrote:

> @willmot <https://github.com/willmot> can you remind me about this?
>
> Make sure we are passing in the same args as we pass to mysqldump.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/humanmade/backupwordpress/pull/835#issuecomment-150177953>
> .
>



-- 
-- Use of a keyboard or mouse may be linked to serious injuries or
disorders.
diego dot torres at gmail dot com - Madrid / Spain

willmot:
@pdewouters as we're using both `mysqldump` via cli and `php-mysqldump` as a fallback we should make sure we have them both congfigured the same with the same options passed etc.

pdewouters:
thanks @ifsnop for the update

willmot:
All tests pass, merge train is leaving the station

🚄 choo choo

pdewouters:
@ifsnop did you ever release an update for the bugfix you mention?

ifsnop:
Yes, it was done a few hours after my post, you can see it here https://github.com/ifsnop/mysqldump-php/releases/tag/v2.1

pdewouters:
thanks!

PR #841, opened 2015-07-18
Use the new backupwordpress@hmn.md email address everywhere

willmot:
This PR switched out all instances of `hello@` and `support@` for our new `backupwordpress@` email address.

Doing this also affected most of the translation `.po` files as those have email addresses in them, thus I reserved and regenerated the `.mo` as well. Because of that there are also some changes to the metadata and line endings in some of the translation files.


Gruntfile.js
```diff
@@ -18,7 +18,7 @@ module.exports = function (grunt) {
 					mainFile  : 'backupwordpress.php',         // Main project file.
 					type      : 'wp-plugin',    // Type of project (wp-plugin or wp-theme).
 					processPot: function( pot, options ) {
-						pot.headers['report-msgid-bugs-to'] = 'support@humanmade.co.uk';
+						pot.headers['report-msgid-bugs-to'] = 'backupwordpress@hmn.md';
```
willmot: Wat, where does this come from. Also the line is shorter than the one it replaces...

pdewouters: lol, I had added this integration a while back! https://houndci.com/

willmot: ah hahaha

PR #653, opened 2015-07-18
Drop support for PclZip

pdewouters:
It would be good to get data on what % of users don't have `zip` or `zipArchive` and thus have to fall all the way back to PclZip. Supporting PclZip is a PITA so if we can drop it I'd be +1


pdewouters:
Any reason we're using pclzip here https://github.com/humanmade/backupwordpress/blob/master/tests/class-wp-test-hm-backup-testcase.php#L15

pdewouters:
The whole BackUp class was indented inside a `namespace` block, that was removed and consequently, the indentation for the whole class changed

willmot:
> I don't know if that definition of an active user applies to us, but it does give an idea of proportions.

Of the 6,906 people that have contacted support, only 1.75% relied on PclZip. If we segment out only active users then that drops to 0.40%, so it seems the trend is that fewer people need it over time.

Extrapolating out to our supposed 200,000 that's still around 3,500 (worst case) to 800 users in total which whilst not nothing, isn't a huge number.

> Any reason we're using pclzip here

The unit tests work by doing a backup and then comparing the contents of the backup to the list of files that should be in there. Those tests could be refactored to use `ZipArchive` but it doesn't have an equivalent method to extract the list of files to an array so we'd likely need to do some refactoring to instead extract to a tmp directory prior to comparing.

> The whole BackUp class was indented inside a `namespace` block, that was removed and consequently, the indentation for the whole class changed 

`?w=1` FTW 🚀 https://github.com/humanmade/backupwordpress/pull/653/files?w=1

pdewouters:
> ?w=1 FTW

Ah yes forgot about that, handy!

pdewouters:
> Those tests could be refactored to use ZipArchive
This can be a spearate issue

willmot:
Discussed in [Slack](https://backupwordpress.slack.com/archives/general/p1437746625000190)

- We're going to keep `PclZip` around for unit tests for now
- We need to audit to see if we've removed all the code we can
- We need to audit copy / documentation to make sure we've removed mention of it.

We're going to drop support for PclZip!

willmot:
On my list to review/merge

willmot:
@owaincuvelier looks like we just need to handle documentation updates before we ship this. Could you do an audit of what we'll need to update?

owaincuvelier:
@willmot yeah no problem. I'll get on it early next week.

pdewouters:
I don't think that's a blocker for merging this PR though?

willmot:
@pdewouters agreed, here comes the merge train, choo, choo 🚄🙀🚄

owaincuvelier:
rather then remove it I might change references to a deprecated notice for now, and completely remove after we hit higher adoption of the latest version.

willmot:
@owaincuvelier great idea

PR #840, opened 2015-07-18
Separate Github and wp.org readme's

pdewouters:
Security issues are unavoidable, when people discover them we should help them disclose the issue to us responsibly.

- [x] Add a section to reporting security issues to the [contributing guidelines](https://github.com/humanmade/backupwordpress/blob/master/CONTRIBUTING.md)
- Do we need a wiki page with a deeper explanation, probably not.


pdewouters:
As mentioned, we can stop syncing the WordPress readme with the Github readme.
How about taking the same approach as the WP API for reporting security issues? I opened a hackerone account for this if we want to do it.

willmot:
> As mentioned, we can stop syncing the WordPress readme with the Github readme.

Great, sounds like a good approach, I wonder how we should separate concerns between `contributing.md` and `readme.md`? Readme has the benefit of being shown on the main page, but, the contributing guidelines are highlighted on the new issues page. should we add a section on reporting security issues to the contributing.md file as well?

> How about taking the same approach as the WP API for reporting security issues? I opened a hackerone account for this if we want to do it.

Yeah I think that's a good idea, @joehoyle do we already have a Hackerone account, is it a good approach? Should we have a central HM account or one per product?

> The dev blog URL is not definitive, https://dev.bwp.hmn.md

Seems like a good url 👍

joehoyle:
We don't have a hacker one account, and I'm not sure how much of a good idea it is. If we have a security reporting channel that might be good enough. If we start offering bounties on hacker one we'll probably get a lot of invalid inbounds to sift through. I'd say having an easy to discover security email would be good enough.

rmccue:
From experience, there's a lot of crap reports through HackerOne, and they don't really make up for the actual reports you get through it.

willmot:
Cool lets skip hackerone and just add to `contributing.md` and `readme.md`

willmot:
@pdewouters I think we should use this PR to just add security copy to contributing file, switching the readme warrants it's own issue I think (where we can discuss what goes in it, I like https://github.com/humanmade/cavalcade as an example).

pdewouters:
- [ ] remove hackerone mention, we are not using it

willmot:
- [ ] Move security section to contributing.md file and just link to it from readme.

willmot:
Fixed up the contributing.md file in https://github.com/humanmade/backupwordpress/issues/871.

Going to rename this issue to reflect the changes

willmot:
@pdewouters fixed up the language files that were accidentally removed.

PR #869, opened 2015-09-08
Inspection fixes

pdewouters:


pdewouters:
How weird, that it's failing on the master branch only

pdewouters:
This unit test is failing on the `master` branch of WordPress

````
	public function testIsPathProtected() {

		// Fake that we're on Apache so we can also test .htaccess
		global $is_apache;
		$is_apache = true;

		// Test the default backup path
		$this->assertFileExists( $this->path->get_path() . '/index.html' );
		$this->assertFileExists( $this->path->get_path() . '/.htaccess' );

		// Test a custom backup path
		$this->path->set_path( $this->custom_path );

		$this->path->calculate_path();

		$this->assertFileExists( $this->path->get_path() . '/index.html' );
		$this->assertFileExists( $this->path->get_path() . '/.htaccess' );

	}
````

File: backupwordpress.php [hunk 1]
```diff
@@ -6,7 +6,7 @@
 Version: 3.3 Alpha
 Author: Human Made Limited
 Author URI: http://hmn.md/
-License: GPL-2.0+
+License: GPL-3.0
```
willmot: Was this intentional? GPL 2.0+ and GPL 3.0+ are different licenses.

pdewouters: `GPL-2.0+` - I assum means 2 and higher?

From https://wordpress.org/about/license/

> The license under which the WordPress software is released is the GPLv2 (or later)

I started looking at the differences but it gave me a headache.

I got a notice in Scrutinizer stating that it was not a valid SPDX license. So I checked out that page https://spdx.org/licenses/

I'm fine with reverting it back.

willmot: JetPack simply goes with `GPL2+`: https://github.com/Automattic/jetpack/blob/master/jetpack.php#L10, I vote we switch to that.

pdewouters:
@willmot fixed as per discussion

PR #873, opened 2015-09-25
Site size isn't ignoring excluded files

willmot:
And it should be, this was broken at some point. I think in https://github.com/humanmade/backupwordpress/pull/743

I found two issues.

- We were no longer passing the `$skip_excluded_files` argument all the way through to `filesize`. Fixed that in https://github.com/humanmade/backupwordpress/commit/16d0ddfb8485e2a3dc0a59bc6d34f172078bb526
- https://github.com/humanmade/backupwordpress/commit/d8ba9e4f15ccfdc480b03ee51cc2fbf16966a6e8 shortcuts filesize so it doesn't ever get to the exclude ignoring code if it's passed the route directoy. @pdewouters any idea why that was added?

### DIFF
--- a/admin/schedule-sentence.php
+++ b/admin/schedule-sentence.php
@@ -155,7 +155,7 @@ function hmbkp_get_site_size_text( HM\BackUpWordPress\Scheduled_Backup $schedule
 
 	} elseif ( ( 'database' === $schedule->get_type() ) || $schedule->is_site_size_cached() ) {
 
-		return sprintf( '(<code title="' . __( 'Backups will be compressed and should be smaller than this.', 'backupwordpress' ) . '">%s</code>)', esc_attr( $schedule->get_formatted_site_size() ) );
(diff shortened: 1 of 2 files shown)


willmot:
Here's a screenshot of the issue, the site size should be taking into account that the `wordpress` and `contents` directories are ignored.

screen shot 2015-09-25 at 15 56 50

willmot:
We could consider implementing the improvements detailed here: https://github.com/humanmade/backupwordpress/pull/743#issuecomment-76792740

pdewouters:
@willmot want to release this as part of 3.3? We can have a separate issue for any enhancements.

willmot:
@pdewouters yep let's do that 👍

owaincuvelier:
screen shot 2015-11-09 at 16 41 59

@pdewouters have we released this yet?

pdewouters:
@owaincuvelier yes, but we found another issue. https://github.com/humanmade/backupwordpress/pull/932

owaincuvelier:
@pdewouters oh whoops. Thanks for pointing that out!

Issue #860, opened 2015-08-04
Consider using wp_normalize_path


willmot:
https://core.trac.wordpress.org/ticket/33265

willmot:
Closing in favor of #874

PR #874, opened 2015-09-25
Switch to wp_normalise_path

willmot:
And drop our `Backup::conform_dir` method, see https://github.com/humanmade/backupwordpress/issues/860

Our original `conform_dir` unit tests all still passed when updated to use `wp_normalise_path`, with the exception of `testConformDirTestCase::testTrailingSlash`.

The behavious of `wp_normalize_path` differs from `conform_dir` when it comes to trailing slashes. We removed them but core leaves them be, thus our unit test for a path with a trailing slash fails. We're **not** actually relying on that behaviour anywhere so it doesn't actually cause any issues. Also looking into convention it seems that paths are generally better off having a trailing slash rather than not; for two reasons:

- a trailing slash is a nice indicator that a path is a directory is a path vs a file.
- having a double slashes in a path doesn't cause an error whereas ommitting the slash fully obviously does.

So we should switch to `wp_normalize_path` as is and remove our unit tests completely.


pdewouters:
nice, that function is there since 3.9.0!

Issue #789, opened 2015-04-05
Deprecated function size()


pdewouters:
The `size()` function has been deprecated as of jQuery 1.8
https://api.jquery.com/size/

We should replace all instances with the `length` property

willmot:
@pdewouters is this in JS? Still relevant?

PR #875, opened 2015-09-25
use .length instead of .size() as it's deprecated since jQuery 1.8

willmot:
Fixes https://github.com/humanmade/backupwordpress/issues/789


assets/hmbkp.js
```diff
@@ -125,7 +124,7 @@ function hmbkpToggleScheduleFields( recurrence ) {
 
     var settingFields = jQuery( '.recurring-setting' );
     var scheduleSettingFields = jQuery( '#schedule-start' );
-    var twiceDailyNote = jQuery( 'p.twice-js' );
+    var twiceDailyNote = jQuery( '.twice-js' );
```
willmot: This change should actually have been part of https://github.com/humanmade/backupwordpress/commit/0abea487e40f10ae4e3155c28ce9b0a7c2c4ca7, I'll leave in here as I assume this is going to be merged before the next release anyway.

pdewouters:
no need to wait for build to finish as these are JS changes

PR #880, opened 2015-09-26
Fix wording and typos

pdewouters:


pdewouters:
No need for CI to finish here as it only touches text

PR #883, opened 2015-09-26
Update WordPress versions

pdewouters:


willmot:
@pdewouters if we're going to stop testing on 3.9 and 4.0 should we bump the minimum plugin version to 4.1?

pdewouters:
Good question, it still works with 3.9 though. Not on earlier versions because `wp_normalize_path` requires 3.9

willmot:
I don't think we should remove the automated testing until we bump the minimum version. If we're going to bump the minimum version it would be good to get some stats.

pdewouters:
Here's the breakdown for WordPress versions https://wordpress.org/about/stats/
Looks like 3.9 is only 6%

pdewouters:
Ah, we're not tracking the WordPress version

willmot:
waaaat, let's start tracking it :+1:

pdewouters:
So what would you do in the meantime? Testing 4 versions is going to take a long time, as the build is already quite slow. But I'd like to start testing the latest versions now

willmot:
@pdewouters It would be good if we can solve the speed issue without having to give up testing older versions which we still support.

Are there other angles for speeding thngs up that we can take before we do this?

willmot:
@pdewouters assume this can be closed off now?

PR #889, opened 2015-09-29
Maximum warp

pdewouters:
- test in HHVM and PHP 7 ( allowing failures )
- test 3 most recent version of WP and earliest supported
- do not do Codesniffer tests here as it is handled by Scrutinizer
- Use the git version of the WordPress trunk repo ( config from WP API repo )
- only run code coverage on latest PHP/WP

**Before**
screenshot 2015-09-30 13 43 18

**After**
screenshot 2015-09-30 13 43 06


.travis.yml
```diff
@@ -1,39 +1,88 @@
...
+  - vendor
+  - $HOME/.composer/cache
+
+before_install:
+  # set up WP install
+  - export WP_DEVELOP_DIR=/tmp/wordpress/
+  - mkdir -p $WP_DEVELOP_DIR
+  - git clone --depth=1 git://develop.git.wordpress.org/ $WP_DEVELOP_DIR
+  # set up tests config
+  - cd $WP_DEVELOP_DIR
```
willmot: Don't we also need to checkout the correct version tag here so that we're testing against the version of WordPress in `WP_VERSION`? In the current config I don't see how Travis would know to get the version of WordPress we're supposed to be testing against?

willmot:
Amazing work @pdewouters, thats a 663% improvement!

Couple of thoughts:

- [ ] Given it's pre-release can we just run the PHP7 and HHVM tests against the latest WP Release instead of once for each release?
- [ ] Given single site is so much more common would it be better test all single versions but only the latest version with multisite?
- [ ] Are these notices avoidable?
```
The command "if [[ "$TRAVIS_PHP_VERSION" == "5.6" && "$WP_VERSION" == "latest" ]] ; then
    phpunit --testsuite backup --coverage-clover=coverage.clover
    phpunit --testsuite schedule --coverage-clover=coverage.clover
    phpunit --testsuite other --coverage-clover=coverage.clover
else
    phpunit --testsuite backup
    phpunit --testsuite schedule
    phpunit --testsuite other
fi
" exited with 0.
```
- [ ] Would it be quicker to run all our tests at once rather than splitting over 3 runs?
- [ ] Any reason we need `depth=50` instead of `depth=1` when cloning BackUpWordPress? `git clone --depth=50 https://github.com/humanmade/backupwordpress.git`

pdewouters:
> Are these notices avoidable?
which notices?

willmot:
The `" exited with 0.` blocks. Or am I getting confused and they're not notices?

pdewouters:
Ah the exit code. Aren't they useful for debugging? Not sure they can be silenced

pdewouters:
@willmot implemented your suggestions. Looks like it still runs ok, and faster as we are running less jobs.

pdewouters:
The `excludes` section is verbose. Would be nice to figure out a more succint way to define the matrix

willmot:
Woop woop!
screen shot 2015-09-30 at 16 15 29

willmot:
Fantastic work @pdewouters!

Issue #898, opened 2015-10-02
Force zip method condition will never apply if zip is available


pdewouters:
This conditional here: https://github.com/humanmade/backupwordpress/blob/master/classes/class-backup.php#L921

If `HMBKP_FORCE_ZIP_METHOD` is set to `ziparchive`, but `get_zip_command_path()` is true, it will still use `zip` instead of forcing `ZipArchive`

We should check the constants first.

PR #901, opened 2015-10-03
Respect constant override

pdewouters:
Fixes https://github.com/humanmade/backupwordpress/issues/898


classes/class-backup.php
```diff
@@ -917,15 +917,33 @@ public function mysqldump_fallback() {
...
-			$this->zip_archive();
+		if ( defined( 'HMBKP_FORCE_ZIP_METHOD' ) ) {
+			switch ( HMBKP_FORCE_ZIP_METHOD ) {
+				case 'zip':
+					if ( $this->get_zip_command_path() ) {
+						$this->zip();
+					}
+					break;
+				case 'ziparchive':
+					if ( class_exists( 'ZipArchive' ) ) {
```
willmot: Need to handle `class_exists( 'ZipArchive` )` being false.

classes/class-backup.php
```diff
@@ -917,15 +917,37 @@ public function mysqldump_fallback() {
...
+					break;
+			}
 		} else {
-			$this->warning( $this->get_archive_method(), __( 'No valid archive method found.', 'backupwordpress' ) );
+			// Is zip available
+			if ( $this->get_zip_command_path() ) {
+				$this->zip();
+			}
+			// If not or if the shell zip failed then use ZipArchive
+			elseif ( empty( $this->archive_verified ) && class_exists( 'ZipArchive' ) && empty( $this->skip_zip_archive ) ) {
```
willmot: Now that we don't support PclZip I think we can drop `skip_zip_archive` completely?

Also I think there's an issue here with using `elseif`, we want to fallback to `ZipArchive` if `zip` doesn't work for whatever reason, even if `zip_command_path` is valid. Currently we'll never fallback to `ZipArchive` because of the `elseif`.

pdewouters:
oops, broke the tests

pdewouters:
@willmot if you're happy with this, we can release 3.3

tests/backup/testBackUpProcessTest.php
```diff
@@ -1,4 +1,4 @@
-<?php
```
willmot: Accidental tab here?

Issue #525, opened 2014-08-19
One click installation of extensions


owaincuvelier:
It is possible to use the Stripe AP and EDD to enable one click installation of extensions from the WordPress dashboard. It was demonstrated by Thomas Griffin (can't find the link though), this would surely increase conversions from the free BackUpWordPress plugin.

I'm not sure how this would work with EDD though. But it's a cool idea

pdewouters:
Would be nice to replace the upsell links with "install" buttons. We could mimick the plugins search results design and use the same ajax install process.clicking install would popup a modal, or slide-down a purchase form.
we may be able to see thie imlementation as an EDD addon.

pdewouters:
So this is the flow I had in mind

- user visits addons catalog page
- User clicks "install"
- Display payment form
- On success install addon from remote location
- Successful payment should trigger an EDD entry and send license to user
- Install successful - prompt user for license key.
- finished

We would need to be able to use the Software Licensing addon API to trigger a new purchase if payment is successful

https://easydigitaldownloads.com/support/topic/create-purchase-via-api/

willmot:
It would be nice if it also automatically entered the correct license, that way the process could be completely seamless / "it just works"

pdewouters:
yes, it would. I suppose if the API returns the license on a successful request.

pdewouters:
the EDD API doesn't support creating license keys, so we'd still need to send an email with the key for then to enter manually.

pdewouters:
Relevant code example https://github.com/humanmade/hm-stream/blob/f77afa5515daf3b0aa8c0226343812816271e3ea/includes/extensions.php

willmot:
The first bash at this is implemented in #909. We'll leave this open to cover future enhancements, including moving the purchasing flow inline.

libby-barker:
@willmot and @pdewouters what work remains on this one?

willmot:
We have an extensions page within the plugin. However clicking install on one of those extensions just links the user out to the bwp.hmn.md website where they then have to complete the normal purchase flow, receive an email with links to download the extension plugins, download those plugins, upload them to their site, activate them and then paste in the correct API key.

Instead we should work to bring that entire process inline as part of the plugin. The flow outlined in https://github.com/humanmade/backupwordpress/issues/525#issuecomment-81619336 is still relevant. Although I think we could also auto-populate the license key.

Issue #762, opened 2015-03-12
Use the Notices class throughout


pdewouters:
There are several methods we use to bubble up errors to the user, mainly using the Notices class, the HM-Backup error() method and `hmbkp_get_settings_errors`
We should use one consistent method.
Also, I think we can take the opportunity to review how each function handles errors, it seems inconsistent. We should either always return a `WP_Error` or an `Exception`

PR #909, opened 2015-10-09
Add the beginnings of a dedicated page for extensions.

willmot:
Starts to lay the ground-work for one-click extension installation #525 although this could (and probably should) be released without that.

This copies the layout and design from the install plugins screen and adds a link to the page title on the Backups page:
screen shot 2015-10-09 at 16 50 17
screen shot 2015-10-09 at 16 50 09

For now we're using dummy data.

Todo

- [x] Should pull extension info from bwp.hmn.md
- [x] The page shouldn't show in the tools menu as a separate menu item, it should only be accessable from the link in the main page H1.
- [x] For now the "Install" button could just take people to the page on bwp.hmn.md
- [x] Highlight the new button with a feature pointer
- [x] Decide what to do with the existing links to the extensions in the footer of the backups page.
- [x] Translate the words


pdewouters:
Can we use the WP API with endpoints for the Downloads CPT?

willmot:
> Can we use the WP API with endpoints for the Downloads CPT?

Awwwww yeah!

pdewouters:
> EDD 2.5 is going to introduce a new class (EDD_Payment) that will make generating the new payment record and license keys substantially easier.

2.5 is probably a month away.

willmot:
Chatted to Paul and he's going to take this over from me.

pdewouters:
@willmot where can I get those red icons? are there different sizes available?

willmot:
@pdewouters I don't think the source files exist anymore unfortunately, we only have what's available in the theme to work with.

pdewouters:
it's a lot of code to add a pointer! https://gist.github.com/brasofilo/6947539

pdewouters:
The live site API response does not contain `featured_image_thumbnail_url` , but my local does.

pdewouters:
> Decide what to do with the existing links to the extensions in the footer of the backups page.

How about just changing it to a single link: "Store your backups securely in a remote location with our extensions"

willmot:
> How about just changing it to a single link: "Store your backups securely in a remote location with our extensions"

That sounds good, although I'd go with "in the Cloud" instead of "in a remote location".

pdewouters:
yeah, just need to make sure my cloud to butt extension is disabled :D

richardswain:
@pdewouters your butt cloud extension has a place in history

pdewouters:
ha, reading that back, "a remote location" does sound funny :)

pdewouters:
Looks pretty much the same, just missing the images.
screenshot 2016-03-04 15 35 16

pdewouters:
It would be good to have some `128px` icons because these are `60px` so they are a bit pixelated.

willmot:
Took a look at this, some things I think we need to improve before launch:

- [ ] Show the icons at their native size as otherwise they look very pixilated.
- [ ] Append ' ago' to the "Last Updated" string to match Core's Add Plugins page.
screen shot 2016-03-07 at 12 33 23
- [ ] Can we move the Developer Bundle to first and highlight in someway?
- [ ] Can we show the latest plugin version left aligned where we show "Last Updated"?
- [ ] We need an icon for the developer bundle.
- [ ] Can we still highlight the "backups" link in the wp-admin menu when on the extensions page?
- [ ] Let's remove the "by Human Made Limited" line for now, they're all by us so is redundant.
- [ ] typo, `providor` should be `provider`.
- [ ] Need to do a copy pass across each of the extensions as it's pretty wildly inconsistent. @willmot to tackle.

pdewouters:
screenshot 2016-03-14 15 13 11

Logos are now 512x512

pdewouters:
Added a box logo for the bundle, the Win Azure and S3 are a bit different, it was difficult to find the right size imags

willmot:
@pdewouters those look loads better!

pdewouters:
to get the bundle in the first position, we could order the posts alphabetically, but this doesn't seem to work 
`https://bwp.hmn.md/wp-json/wp/v2/edd-downloads?orderby=title&order=desc`

pdewouters:
Used different background color for the footer of the bundle
screenshot 2016-03-15 10 03 57

Too subtle?

pdewouters:
> Can we still highlight the "backups" link in the wp-admin menu when on the extensions page?
Could this cause confusion? The only way I can think of is using JS?

willmot:
> Could this cause confusion? The only way I can think of is using JS?

I don't think so, we're in the backups "section" after-all. For example the categories link stay highlighted on both the main categories screen and when editing a single category, similarly for media when editing a single item.

> Too subtle?

I think that's good enough for now, we can always tweak / change later 👍

admin/actions.php
```diff
@@ -653,3 +653,11 @@ function ajax_cron_test() {
 
 }
 add_action( 'wp_ajax_hmbkp_cron_test', 'HM\BackUpWordPress\ajax_cron_test' );
+
+/**
+ * Remember notice dismissal
+ */
+function hmbkp_dismiss_notice() {
```
willmot: Could we extend `Notices` to cover this use case?

admin/extensions.php
```diff
@@ -0,0 +1,95 @@
...
+
+namespace HM\BackUpWordPress;
+
+?>
+
+<div class="wrap">
+
+	<h1>
+
+		<a class="page-title-action" href="<?php echo get_settings_url() ?>"><?php _e( '&larr; Backups', 'backupwordpress' ); ?></a>
```
willmot: `get_settings_url` would ideally be escaped.

admin/extensions.php
```diff
@@ -0,0 +1,95 @@
...
+									<img src="<?php echo esc_url( $extension->featured_image_url ); ?>" class="plugin-icon" alt="">
+								</a>
+							</h4>
+						</div>
+
+						<div class="action-links">
+							<ul class="plugin-action-buttons">
+								<li>
+									<?php if ( in_array( strtolower( $extension->title->rendered ), $installed_plugins ) ) : ?>
+										<span class="button button-disabled" title="<?php _e( 'This extension is already installed and is up to date', 'backupwordpress' ); ?>"><?php _e( 'Installed', 'backupwordpress' ); ?></span>
```
willmot: >  and is up to date

Is that correct? I don't see any code dealing with checking whether the Extension is in-fact up-to-date, perhaps just legacy text copied from plugins page?

admin/extensions.php
```diff
@@ -0,0 +1,95 @@
...
+									<?php else : ?>
+										<a class="install-now button-primary" data-slug="<?php echo esc_attr( $extension->slug ); ?>" href="<?php echo esc_url( $extension->link ); ?>" aria-label="Install <?php echo esc_attr( $extension->title->rendered ); ?> now" data-name="<?php echo esc_attr( $extension->title->rendered ); ?>"><?php _e( 'Buy Now &dollar;', 'backupwordpress' ); ?><?php echo esc_html( $extension->edd_price ); ?></a>
```
willmot: I think the price should be part of the translatable string here so that translators can move it around as needed as in some languages the currency symbol comes after the price.

admin/page.php
```diff
@@ -7,9 +7,13 @@
 <div class="wrap">
 
 	<h1>
+
 		BackUpWordPress
 
+		<a class="page-title-action" href="<?php echo get_settings_url( HMBKP_PLUGIN_SLUG . '_extensions' ); ?>">Extensions</a>
```
willmot: `get_settings_url` should be escaped.

classes/class-extensions.php
```diff
@@ -0,0 +1,59 @@
...
+
+		if ( is_wp_error( $response ) || empty( $response['body'] ) ) {
+			return false;
+		}
+
+		return json_decode( $response['body'] );
+
+	}
+
+	function fetch( $endpoint, $ttl = 10 ) {
```
willmot: Should we consider caching the data for much longer than this, say 24 hours? Otherwise the data will essentially refresh every time someone visits the page which could be slow.

classes/class-services.php
```diff
@@ -69,6 +69,10 @@ public static function get_services( Scheduled_Backup $schedule = null ) {
 
 	}
 
+	public function get_available_services() {
```
willmot: Looks like this can be removed.

classes/class-extensions.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace HM\BackUpWordPress;
+
+class Extensions {
```
willmot: Ideally we'd:

- Docblock the class and it's methods
- Add unit tests for it's functionality
- Follow the Singleton pattern set by `Path` and `Notices`

willmot:
Would also be good to run through the Scrutiniser issues: https://scrutinizer-ci.com/g/humanmade/backupwordpress/inspections/f96d2b5c-7645-42b9-a4b4-ba30f46d7576 that this introduces.

admin/actions.php
pdewouters: yes, I thought about this. but in this PR?

classes/class-extensions.php
pdewouters: thanks, oversight

admin/extensions.php
pdewouters: screenshot 2016-03-15 14 20 11

Now we detect if an update is required and link to the updates page

pdewouters:
Not too sure of the usefulness of the test for the API response. Do we want to test the contents of the response? Then we shouldn't mock the reponse. See `Tests_External_HTTP_Basic` in core tests.

Or do we simply want to test the `get_edd_data()` function returns an array as expected.

Do we want to test the lower level `fetch()` function, check if it sets a transient,...

admin/actions.php
willmot: Happy for that to be a separate issue, perhaps even part of https://github.com/humanmade/backupwordpress/issues/762

classes/class-extensions.php
```diff
@@ -0,0 +1,103 @@
...
+
+		$cache_key = md5( $request_url );
+
+		$cached = get_transient( 'bwp_' . $cache_key );
+
+		if ( $cached ) {
+			return $cached;
+		}
+
+		$response = wp_remote_get( $request_url );
```
willmot: Should we do some error handling here, what happens if we get a `WP_Error`, or bwp.hmn.md is down for some reason?

classes/class-plugin.php
```diff
@@ -415,6 +425,38 @@ public function load_intercom_script() {
...
+
+		$page = is_multisite() ? HMBKP_ADMIN_PAGE . '-network' : HMBKP_ADMIN_PAGE;
+		if ( $current_screen->id !== $page ) {
+			return;
+		}
+
+		/* translators: %1$s and %2$s expand to anchor tags linking to the new extensions page. */
+		$info_message = sprintf(
+			__( 'Thanks for updating BackUpWordPress, why not check out %1$sour extensions?%2$s', 'backupwordpress' ),
+			'<a href="' . get_settings_url( HMBKP_PLUGIN_SLUG . '_extensions' ) . '">',
```
willmot: `get_settings_url` should be escaped.

classes/class-plugin.php
```diff
@@ -415,6 +425,38 @@ public function load_intercom_script() {
...
+			'</a>'
+		);
+		?>
+
+		<div id="hmbkp-info-message" class="updated notice is-dismissible">
+
+			<p><?php echo $info_message; ?></p>
```
willmot: Ideally would be escaped with `wp_kses`

classes/class-extensions.php
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**
+	 * Parses the body of the API response and returns it.
+	 *
+	 * @return array|bool|mixed|object
+	 */
+	public function get_edd_data() {
+
+		$response = $this->fetch( 'edd-downloads', 600 );
```
willmot: Is the `600` ttl intentional here? It's overriding the default of `DAY_IN_SECONDS`.

willmot:
> Need to do a copy pass across each of the extensions as it's pretty wildly inconsistent. @willmot to tackle.

I've moved this to https://github.com/humanmade/Human-Made-website/issues/812 as no need to hold up this for it.

willmot:
> Not too sure of the usefulness of the test for the API response.

Yeah agreed, what you have looks like a good start, we can add more as issues / changes come in.

classes/class-extensions.php
pdewouters: oh snap

classes/class-extensions.php
pdewouters: the errors are handled in the `get_edd_data` function

willmot: Do we want to store errors in the transient though?

pdewouters: but might be worth checking for response code 200

PR #912, opened 2015-10-09
Beginnings of a welcome process

willmot:
Introduces the conecept of a first-run page.

This will be an entirely new page, modeled on the core about page

The first run tasks will be broken into steps, each step represents something the user should do on the first run. Steps are marked with a tick when they are complete and are worked through in order, steps are collapsed showing only the title and icon, the current step is then expanded.

The whole process can be skipped which just takes people straight to the backups page.

When it's completed the user is taken to the backups page.

screen shot 2015-10-09 at 18 07 39


pdewouters:
Gravity forms has a welcome wizard : 
http://recordit.co/WrZAsMP8Zs

willmot:
Going to close this out, there really isn't anything done here and it's super out of date now.

PR #927, opened 2015-10-23
Beginning of catching fatal errors and timeouts

pdewouters:
### Report fatal errors and fail gracefully.

- Manual jobs
  - [ ] need to reload the admin and show the cross icon
  - [ ] **time outs:** JS polling for manually started jobs. remove heartbeat 
- Scheduled ( cron ) jobs

Introduce a task manager singleton class which manages an array of Backdrop tasks ( created by running a schedule ). We need this to delete the backdrop task on error.

- [ ] amend unit tests
- [ ] handle cases where 1 schedule has multiple tasks. ( duplicate key )

### Timeouts

For manual jobs, we need to do some JS polling of the backup process to notify browser. 
For cron jobs, we need to send an email on failure.

In both cases, we can use the status file `.backuprunning` to check last status start time. If > 30min then we can decide it has failed. 
Or do we want to make it more responsive?, we'd have to continuously be writing/reading a status for that.

### DIFF
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -79,6 +79,13 @@ function hmbkp_request_do_backup() {
 
 	$schedule_id = sanitize_text_field( urldecode( $_POST['hmbkp_schedule_id'] ) );
 
+	// If a backdrop task already exists for this schedule, delete it.
+	$task = \HM\BackUpWordPress\Task_Manager::get_instance()->get_task( $schedule_id );
+
+	if ( ! is_wp_error( $task ) ) {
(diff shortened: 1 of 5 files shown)


willmot:
nice!

pdewouters:
So, when we first check the backup status from JS, the file hasn't even been created yet so we exit the polling function. there needs to be a way to detect this case.

willmot:
Going to close this in favour of https://github.com/humanmade/backupwordpress/pull/1025.

PR #932, opened 2015-10-24
Fix an issue that caused the site size to report as twice as large as it should

### LINKED ISSUE #930: Excluding files should not trigger the filesize scanner to re-run
I noticed this happening on a client site, needs confirming.

willmot:
Don't special case pre-calculate the root size during the directory scanner, instead rely on the directory `array_sum` like any other path.

Fixes #930


willmot:
@pdewouters this fixes one of the issues we saw debugging the support issue the other day.

pdewouters:
confirmed

PR #940, opened 2015-11-13
Support modal broken

owaincuvelier:
After you enable support, the support modal is now broken.


owaincuvelier:
@pdewouters this fixes the intercom modal, just added the ID back to the element. Can you merge? Be good if we could release this fix ASAP.

PR #951, opened 2015-12-10
Ensure row actions always show

willmot:
4.4 hides `.row-actions` by default with the addition of:

```CSS
position: relative;
left: -9999px;
```

This PR overrides that by setting the `position` back to `static` so our backup action links show as they should.

screen shot 2015-12-10 at 12 57 03


willmot:
Thanks @pdewouters

PR #963, opened 2016-01-21
Avoid a warning by ensuring we don't pass a string to array_filter

willmot:


willmot:
@pdewouters Have fixed the root cause and removed the need for array_filter etc.

Issue #964, opened 2016-01-21
ERROR on Update to 3.4: open_basedir restriction in effect


katmoody:
From Intercom Ticket: https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1765944325

```
Warning: file_exists(): open_basedir restriction in effect. File(/home/.sites/124/site9073757/wp-config.php) is not within the allowed path(s): (/home/.sites/124/site9073757/web:/home/.sites/124/site9073757/tmp:/usr/share/pear) in /home/.sites/124/site9073757/web/wp-content/plugins/backupwordpress/classes/class-path.php on line 107 Warning: file_exists(): open_basedir restriction in effect. File(/home/.sites/124/site9073757/wp-config.php) is not within the allowed path(s): (/home/.sites/124/site9073757/web:/home/.sites/124/site9073757/tmp:/usr/share/pear) in /home/.sites/124/site9073757/web/wp-content/plugins/backupwordpress/classes/class-path.php on line 112 1 
Warning: file_exists(): open_basedir restriction in effect. File(/home/.sites/124/site9073757/wp-config.php) is not within the allowed path(s): (/home/.sites/124/site9073757/web:/home/.sites/124/site9073757/tmp:/usr/share/pear) in /home/.sites/124/site9073757/web/wp-content/plugins/backupwordpress/classes/class-path.php on line 107 Warning: file_exists(): open_basedir restriction in effect. File(/home/.sites/124/site9073757/wp-config.php) is not within the allowed path(s): (/home/.sites/124/site9073757/web:/home/.sites/124/site9073757/tmp:/usr/share/pear) in /home/.sites/124/site9073757/web/wp-content/plugins/backupwordpress/classes/class-path.php on line 112```

pdewouters:
We had a fix for this, not sure why it doesn't work in this case
https://github.com/humanmade/backupwordpress/blob/11eccaefbdcecc2f400088c344eb38c084efc089/functions/interface.php#L393

willmot:
`is_restricted_custom_path` just tests whether the path itself is restricted, however in [`Path`](https://github.com/humanmade/backupwordpress/blob/master/classes/class-path.php#L107) we now do `file_exists( dirname( $site_path ) . '/wp-config.php' )` to check for the existence of a `wp-config.php` file outside of webroot. That check was added in #936 as a fix for https://github.com/humanmade/backupwordpress/issues/695.

[...]

PR #967, opened 2016-01-23
Improved handling of open_basedir restrictions

willmot:
Check for the existance of `open_basedir` restrictions and ensure we're not accessing paths that fall outside of them.

Removes the need for some error silencing.

I also took the opportunity to improve the error message shown to users who have a restricted custom path.

Fixes https://github.com/humanmade/backupwordpress/issues/964

### DIFF
--- a/classes/class-path.php
+++ b/classes/class-path.php
@@ -103,14 +103,18 @@ class Path {
 
 		$home_path = $site_path;
 
-		// Handle wordpress installed in a subdirectory
-		if ( file_exists( dirname( $site_path ) . '/wp-config.php' ) && ! file_exists( $site_path . '/wp-config.php' ) && file_exists( dirname( $site_path ) . '/index.php' ) ) {
-			$home_path = dirname( $site_path );
-		}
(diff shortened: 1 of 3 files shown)


pdewouters:
tested locally. should we suggest setting the `HMBKP_PATH` constant?

willmot:
@pdewouters I don't follow...

`HMBKP_PATH` defines the path to the directory backups are stored in, this issue is caused by how we calculate the root path of the site `HMBKP_ROOT`. However, the issue doesn't actually cause the site root to be calculated incorrectly, it just causes the warning as a side-effect.

If `open_basedir` is set to the site root (for example `/srv/www/wp.dev/`), in `PATH::calculate_root` we check if `wp-config.php` exists a directory up from site root (`file_exists( '/srv/www/wp-config.php' )`), as this is outside the basedir it causes a warning, that warning is also a valid `false` result though; the `wp-config.php` file can't be outside the basedir as WordPress wouldn't be able to read it either.

pdewouters:
nevermind, I was just thinking about the backups folder being inaccessible..

PR #968, opened 2016-01-23
Show a warning message if neither mysqldump nor PDO are available

### LINKED ISSUE #959: Better error handling when mysqldump and pdo_mysql are both missing
So far, I've gotten a few tickets (and in the forums) in which the update to include MySQLDump actually makes the plugin fail, and it looks like that is because the fallback method isn't working for some reason? 

Examples of tickets: 
Intercom: https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1674279260
wp.org forum: https://wordpress.org/support/topic/mysqldump_fallback

willmot:
We should potentially look at bringin back a non-PDO mysqldump fallback for users without the above.

Fixes #959

### DIFF
--- a/classes/class-requirement.php
+++ b/classes/class-requirement.php
@@ -16,7 +16,7 @@ abstract class Requirement {
 	/**
 	 * @return mixed
 	 */
-	abstract protected function test();
+	abstract public static function test();
 
 	/**
 	 * @return mixed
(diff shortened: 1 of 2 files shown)


classes/class-requirement.php
```diff
@@ -16,7 +16,7 @@
 	/**
 	 * @return mixed
 	 */
-	abstract protected function test();
+	abstract public static function test();
```
pdewouters: I tihnk the abstract one should remain protected?

willmot: Good call.

Issue #972, opened 2016-01-24
Fatal error - call to member function


katmoody:
This person lost access to their site after activating the plugin: 

```
I just installed your plugin and it showed that it was installed successfully. But as soon as I pushed the Activate Plugin link it went to a blank page and gave me this error code:

Fatal error: Call to a member function set_archive_filename() on null in /home/btvsadmin/brandtuitives.com/wp-content/plugins/backupwordpress/functions/core.php on line 110
```

This is on Intercom ticket #https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1777144252

pdewouters:
This is only supposed to run if they are upgrading from an older 1.x version - could you ask if they ever installed the plugin on that site before?

willmot:
Fixed in #974

PR #974, opened 2016-01-25
Ensure backups and schedules are properly migrated from pre-2.0

willmot:
- Fix a fatal error in the upgrade routine
- Ensure old backups are correctly renamed so that the legacy schedule recognises them
- Ensure that the default schedules run after the upgrade routine so that they aren't added if there's already a legacy schedule

Fixes https://github.com/humanmade/backupwordpress/issues/972

### DIFF
--- a/functions/core.php
+++ b/functions/core.php
@@ -107,9 +107,27 @@ function update() {
 		}
 
 		// Set the archive filename to what it used to be
-		$legacy_schedule->backup->set_archive_filename( implode( '-', array( get_bloginfo( 'name' ), 'backup', current_time( 'Y-m-d-H-i-s' ) ) ) . '.zip' );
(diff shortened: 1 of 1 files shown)


willmot:
@pdewouters I verified this by checking out the first tag we have (`1.2`), running a backup and then re-checking out this branch and reloading.

Was was happening is that we got a fatal error. Even after fixing the fatal there were issues:

- We ended up with 3 schedules; a legacy schedule that matched whatever settings they had in 1.2 and then the two default schedules.
- Any backups that had been completed didn't show under the legacy schedule as they weren't named correctly.

Now, when upgrading from 1.2 any old backups are correctly moved to the new backups folder, renamed so the legacy schedule will recognise them, so you end up with a single schedule with the options you had set pre-upgrade with your old backups correctly showing:

screen shot 2016-01-25 at 11 42 35

pdewouters:
nice one

PR #977, opened 2016-01-27
Detect subdir install by comparing home url and site url

### LINKED ISSUE #975: Trying to download backups leading to 404 Error
Multiple users reporting they are unable to click to download the file from their admin. I was able to work on the first site and noted that the download URL doesn't seem to be passed correctly? 

```
http://mefemn.org/site/wp-admin/admin-post.php?hmbkp_backup_archive=L2hlcm1lcy9ib3NvcmF3ZWIwNTMvYjE4NzcvaXB3Lm1lZmVtbm9yZy9wdWJsaWNfaHRtbC9zaXRlL3dwLWNvbnRlbnQvYmFja3Vwd29yZHByZXNzLTM3ZDI4NWRhNTYtYmFja3Vwcy9tZWZlbW4tb3JnLTE0NDAwMTQ5MTItY29tcGxldGUtMjAxNi0wMS0yNi0xNC0wMi0xNy56aXA%3D&hmbkp_schedule_id=1440014912&action=hmbkp_request_download_backup&hmbkp_download_backup_nonce=fa2c67ad27
```

That's what I get if I try and copy the download link so I can just open it in a new window - and it will come back to a 404 error. 
This ticket: https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1774343266

pdewouters:
Should fix #975

### DIFF
--- a/classes/class-path.php
+++ b/classes/class-path.php
@@ -106,9 +106,9 @@ class Path {
 		if ( path_in_php_open_basedir( dirname( $site_path ) ) ) {
 
 			// Handle wordpress installed in a subdirectory
-			if ( file_exists( dirname( $site_path ) . '/wp-config.php' ) && ! file_exists( $site_path . '/wp-config.php' ) && file_exists( dirname( $site_path ) . '/index.php' ) ) {
-				$home_path = dirname( $site_path );
-			} else if ( get_site_url() !== get_home_url() ) {
+			// 1. index.php and wp-config.php found in parent dir
+			// 2. index.php in parent dir, wp-config.php in $site_path ( index.php can be in both locations )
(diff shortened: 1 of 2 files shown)


pdewouters:
@willmot thoughts?

willmot:
@pdewouters this seems like it mightn't be a strict enough check, for example I've run across sites which have two separate domains for `home_url` and `site_url` (for example `site_url` has the hosts temp domain and `home_url` has the actual live domain).

Also we need to update the unit tests to cover this case.

I'm also not 100% why this wasn't already covered as the unit tests should have covered every possible case here.

pdewouters:
@willmot changed function so that `wp-config.php` can be either in parent dir or in subdirectory. Added a test for that case

willmot:
@pdewouters this looks great, nice one.

Issue #978, opened 2016-01-27
Character Encoding Issue Resulting in Missing Images from Backup


katmoody:
Missing picture files have special characters like `ß`, `ä`, `ü`, `ö` in their filename, and the ticket includes multiple examples as well as part of an SQL file with the proper image names. 

In Intercom: https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1748537135

pdewouters:
I wonder if Symfony Finder is causing this issue

katmoody:
Just an update (still on the same ticket there) ... after the update last week, he now has the full backup with images, BUT the slugs have changed. From his newest message: 

```
...now the backup file contains at least all image files (before that update image files with special characters were missing in the backup file).

However, during backup the filenames of images with special characters are changed so that after restore these filenames don’t match with the stored filename/slug in the SQL DB.
```

willmot:
Seems like a perfect chance for some unit tests, we should be able to add a unit test for files with strange characters and have that run across both `zip` and `ZipArchive`.

willmot:
I haven't been able to reproduce this. I've tried two separate approaches.

- I added a unit test to cover backing a file called `Groß.jpg` in https://github.com/humanmade/backupwordpress/pull/981. This passes on both `zip` and `ZipArchive`
- I manually tested by uploading an image called `Groß.jpg`, performing a backup and then completely restoring my site from the backup, the image continued to work fine.

I'm going to test on Windows in-case this is specific to that OS.

willmot:
I wasn't able to reproduce this on Windows either. @katmoody could you point the person who contacted support at this ticket so we can continue debugging.

katmoody:
Done - will update.

katmoody:
Hey - he's updated with a test environment login for you guys - go forth and test merrily :smile:

willmot:
I've had a good look into this and am 99% certain that BackUpWordPress is correct here.

The images have special characters in the filename and those characters are handled correctly when zipping the files up. Similarly the characters are correct in the database backup.

The issue seems to be that when the user has unzipped the files and then transferred them to his server, his file transfer program WinSCP hasn't properly preserved the encoding. Have communicated as much to the original support request.

PR #981, opened 2016-02-02
Unit test backing up files with strange characters

willmot:


willmot:
@pdewouters this just adds a unit test to verify that files with accented characters are correctly backed up. See #978

PR #983, opened 2016-02-04
Ensure our subdirectory check doesn't also match wordpres installed inside wordpress

willmot:
Our subdirectory check was incorrectly matching against the following:

```
index.php // Standard install.
foo/
- index.php // Another standard install in a folder inside another install.
bar/
- index.php // A third WordPress install.

```

In the above case `foo/` & `bar/` would be incorrectly detected as a sub-directory installs, `home_path` would be set to the parent directory and each sub-site would end up backing up the entire root directory.

We resolve this by only matching sub-directory installs if there is an `index.php` in the parent and there isn't one in the current directory.

Also adds a unit test to cover this case.

### DIFF
--- a/classes/class-path.php
+++ b/classes/class-path.php
@@ -105,18 +105,15 @@ class Path {
 
 		if ( path_in_php_open_basedir( dirname( $site_path ) ) ) {
 
-			// Handle wordpress installed in a subdirectory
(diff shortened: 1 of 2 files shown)


willmot:
Closing as this was changed completely in https://github.com/humanmade/backupwordpress/pull/988

Issue #976, opened 2016-01-27
Site Root Path Unreadable


katmoody:
Users are getting a Site Root path unreadable error when they try to scan their site. 
ex. from forums: https://wordpress.org/support/topic/your-site-root-path-isnt-readable?replies=6

Might be related to #975

pdewouters:
This should be fixed by upgrading to 3.4.2

willmot:
@pdewouters don't think this is fixed by 3.4.2, should be fixed by https://github.com/humanmade/backupwordpress/pull/977 though

pdewouters:
@willmot oh, I say that because on my local install I reproduced the error message, being on a 3.4.1 install and as soon as I updated the plugin, the message disappeared

willmot:
Oh hmm, wonder what we changed that fixed that. Looking at https://wordpress.org/support/topic/your-site-root-path-isnt-readable?replies=6 it seems others reported it fixed too. Closing.

katmoody:
Receiving new tickets regarding this today, after updating to 4.3.2 - have logins for troubleshooting but not sure what to troubleshoot myself?

katmoody:
![tools php](https://cloud.githubusercontent.com/assets/5481211/12785492/c29be794-ca5a-11e5-956a-d7b2204d2e37.jpg)
As you can see, in this case the error displays but nothing else loads on the admin page beyond that. 

This is happening on two sites so far within Intercom and both have offered login credentials - I'm just not sure what to check: 

https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1847295244
and 
https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1849907120

pdewouters:
@katmoody I suggest first checking the Intercom data, Looking at this site, it looks like a weird server software is installed. I think we should make it clear that we only support Windows IIS and Apache/nginx on linux

<img width="243" alt="screenshot 2016-02-03 15 45 52" src="https://cloud.githubusercontent.com/assets/30460/12787439/599ab3be-ca8d-11e5-9e22-5159deeb7bd1.png">
<img width="196" alt="screenshot 2016-02-03 15 45 36" src="https://cloud.githubusercontent.com/assets/30460/12787441/599f43c0-ca8d-11e5-941c-1cdd7a19bceb.png">
[...]

PR #988, opened 2016-02-08
Takes logic from core get_home_path

pdewouters:
I extracted the code from the core function `get_home_path` to use in our `get_home_path()` function

Should fix https://github.com/humanmade/backupwordpress/issues/976

### DIFF
--- a/classes/class-path.php
+++ b/classes/class-path.php
@@ -109,13 +109,22 @@ class Path {
 			$siteurl = set_url_scheme( get_option( 'siteurl' ), 'http' );
 			if ( ! empty( $home ) && 0 !== strcasecmp( $home, $siteurl ) ) {
 				$wp_path_rel_to_home = str_ireplace( $home, '', $siteurl ); /* $siteurl - $home */
-				$pos = strripos( str_replace( '\\', '/', $_SERVER['SCRIPT_FILENAME'] ), trailingslashit( $wp_path_rel_to_home ) );
+				$pos = strripos( wp_normalize_path( $_SERVER['SCRIPT_FILENAME'] ), trailingslashit( $wp_path_rel_to_home ) );
 				$home_path = substr( $_SERVER['SCRIPT_FILENAME'], 0, $pos );
 				$home_path = trailingslashit( $home_path );
 			} else {
-				$home_path = ABSPATH;
+				$home_path = wp_normalize_path( ABSPATH );
 			}
 
+			if ( is_multisite() ) {
+				require_once ABSPATH . '/wp-admin/includes/network.php';
+				$hostname          = get_clean_basedomain();
+				$slashed_home      = trailingslashit( get_option( 'home' ) );
+				$base              = parse_url( $slashed_home, PHP_URL_PATH );
+				$document_root_fix = wp_normalize_path( realpath( $_SERVER['DOCUMENT_ROOT'] ) );
+				$abspath_fix       = wp_normalize_path( ABSPATH );
+				$home_path         = 0 === strpos( $abspath_fix, $document_root_fix ) ? $document_root_fix . $base : $home_path;
+			}
 		}
 
 		return wp_normalize_path( untrailingslashit( $home_path ) );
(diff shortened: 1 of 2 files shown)


pdewouters:
looks like we need to change the unit tests

willmot:
We'll need to mock siteurl and homeurl instead

pdewouters:
For multisite, there seems to be some extra checks: https://github.com/WordPress/WordPress/blob/4a6d77c8e5fa846165cae3bb266298f284b43e15/wp-admin/includes/network.php#L334

willmot:
Nice find, seems stupid that isn't rolled into `get_home_path`, but we might as well roll it into ours.

pdewouters:
Here are the core unit tests for that function with some background https://core.trac.wordpress.org/ticket/20449

```
	function test_get_home_path() {
		$home = get_option( 'home' );
		$siteurl = get_option( 'siteurl' );
		$sfn = $_SERVER['SCRIPT_FILENAME'];
		$this->assertEquals( str_replace( '\\', '/', ABSPATH ), get_home_path() );

		update_option( 'home', 'http://localhost' );
		update_option( 'siteurl', 'http://localhost/wp' );

		$_SERVER['SCRIPT_FILENAME'] = 'D:\root\vhosts\site\httpdocs\wp\wp-admin\options-permalink.php';
		$this->assertEquals( 'D:/root/vhosts/site/httpdocs/', get_home_path() );

		$_SERVER['SCRIPT_FILENAME'] = '/Users/foo/public_html/trunk/wp/wp-admin/options-permalink.php';
		$this->assertEquals( '/Users/foo/public_html/trunk/', get_home_path() );

		$_SERVER['SCRIPT_FILENAME'] = 'S:/home/wordpress/trunk/wp/wp-admin/options-permalink.php';
		$this->assertEquals( 'S:/home/wordpress/trunk/', get_home_path() );

		update_option( 'home', $home );
		update_option( 'siteurl', $siteurl );
		$_SERVER['SCRIPT_FILENAME'] = $sfn;
	}
```

pdewouters:
Is it bad practice to require a core file in this way?

pdewouters:
why do we return the home path without the trailing slash?

willmot:
> why do we return the home path without the trailing slash?

There's no good reason, nor is there a general convention either way that I'm aware of, although in WordPress I think only `ABSPATH` includes a trailing slash. I'm actually a fan of always including a trailing slash for reasons details in first answer here: http://stackoverflow.com/questions/17419460/should-i-save-a-path-with-or-without-a-trailing-slash-at-the-end-whats-the-con

pdewouters:
wherever core did `str_replace( '\\', '/', ABSPATH )` , I replaced with `wp_normalize_path`, it seems to do the same thing but with extra checks

pdewouters:
@willmot anything against merging this and releasing a beta?

classes/class-path.php
```diff
@@ -105,18 +105,26 @@ public static function get_home_path( $site_path = ABSPATH ) {
...
-			if ( ( file_exists( dirname( $site_path ) . '/wp-config.php' ) || file_exists( $site_path . '/wp-config.php' ) )  && file_exists( dirname( $site_path ) . '/index.php' ) ) {
-				$home_path = dirname( $site_path );
+			$home    = set_url_scheme( get_option( 'home' ), 'http' );
+			$siteurl = set_url_scheme( get_option( 'siteurl' ), 'http' );
+			if ( ! empty( $home ) && 0 !== strcasecmp( $home, $siteurl ) ) {
+				$wp_path_rel_to_home = str_ireplace( $home, '', $siteurl ); /* $siteurl - $home */
+				$pos = strripos( wp_normalize_path( $_SERVER['SCRIPT_FILENAME'] ), trailingslashit( $wp_path_rel_to_home ) );
+				$home_path = substr( $_SERVER['SCRIPT_FILENAME'], 0, $pos );
+				$home_path = trailingslashit( $home_path );
+			} else {
```
willmot: We don't need this else as `$home_path` is already set to `$site_path` above.

willmot:
@pdewouters looks good, minor note about extraneous `else` statement but other than that happy for you to merge and release a beta 👍

PR #989, opened 2016-02-09
Allow custom secure key using filter

johnnypea:
Sometimes it could be useful to allow custom secure key. I added simple filter for that.

### DIFF
--- a/classes/class-plugin.php
+++ b/classes/class-plugin.php
@@ -289,6 +289,10 @@ final class Plugin {
 	 */
 	protected function generate_key() {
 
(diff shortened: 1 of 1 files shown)


pdewouters:
The WordPress idiomatic way to do it, is to do it at the return point. e.g. `return apply_filters('hmbkp_download_key', md5( serialize( $key ) ));

johnnypea:
I think it depends on context. For example, WordPress uses my method in metadata add/update functions. In this case you want change the whole process. Is it any useful to know the old security key?

P.S.: `hmbkp_generate_key` not `hmbkp_download_key` ;)

Issue #960, opened 2016-01-07
PHP 5.2 non-compat message blocks access to WordPress (front- and back-end)


glueckpress:
![bildschirmfoto 2016-01-07 um 18 07 23](https://cloud.githubusercontent.com/assets/308422/12177365/0a3c9ae8-b56c-11e5-9d3e-f497995096ca.png)

Can’t provide any more details unfortunately. This was an ex-client’s site and their dev meanwhile deleted BackUpWordPress via FTP, so everything is fixed.

However, `wp_die()` blocked access here completely, front- and back-end, so I’d consider this urgent.

willmot:
So the intention of this code is that the plugin shouldn't one able to be activated on < 5.3. That `wp_die` should only ever run inside the activation request where it's then caught and displayed in the plugin activation failed error message.

willmot:
@pdewouters took a look into this and I'm slightly confused by the intention of the code. It seems it's supposed to check for PHP 5.3 in an activation hook, however that never happens as in [`backupwordpress.php`](https://github.com/humanmade/backupwordpress/blob/master/backupwordpress.php) the first thing that happens is that we call `wp_die` if PHP < 5.3.

- We definitely shouldn't be calling `wp_die` as that's going to kill peoples sites. They won't even be able to deactivate the plugin the normal way.
- Instead we should probably just force a deactivation and show an error message.
- But the `Setup` class that defines activation and deactivation and currently has code to self deactivate if PHP is lower than 5.3 **itself requires PHP 5.3+** rendering it useless...

I think the fix here is:

1. Drop the `wp_die` in the main plugin file.
2. Make the `Setup` class PHP 5.2 compatible.
3. Ensure the plugin both won't activate and if already activated, will deactivate itself when it detects PHP 5.2.

pdewouters:
here's the behaviour I see when I try to activate on a PHP 5.2 install, http://recordit.co/SpDEsCjr0z

willmot:
And that message is coming from the `wp_die` statement in `backupwordpress.php` rather than from the activation hook right? So we might as well delete all that code in `Setup` that deals with self-deactivation.

activation hooks don't run on upgrade though right, so if you had an old 5.2 compatible version of the plugin installed and active and then upgraded the plugin would still be active and the whole would be taken down by the `wp_die` call.

pdewouters:
[...]

PR #990, opened 2016-02-11
Avoid killing site if plugin is active on PHP 52

pdewouters:
See #960


willmot:
Ok cool so we're renaming `Setup` to `HMBKP_Setup` so it's PHP 5.2 compatible by removing use of namespace etc.

Seems like a smart approach, did you have opportunity to test on PHP 5.2?

pdewouters:
yes, I used MAMP to switch between versions, when the plugin was activated and it self deactivates correctly.

willmot:
Great, merging then

PR #991, opened 2016-02-19
Ensure we silence any errors from the `Exec` call.

willmot:
Also fixes some minor coding standards violations


willmot:
Still working on this as it looks like there are still issues with PHP WARNINGS being generated when `exec` is disabled.

willmot:
@pdewouters review/merge

This is somewhat related to https://github.com/humanmade/backupwordpress/pull/995 in that it properly catches if the exec call fails (which it will do if for some reason the `is_function_disabled` calls above falsely return true.

willmot:
Closing in favour of #1001

PR #999, opened 2016-02-26
Add a PclZip File Backup Engine

willmot:
Introduces a `Pclzip_File_Backup_Engine` and sets up the common file backup unit tests to run against it.

There are still a couple of issues present that are causing failing tests:
- Broken symlinks cause the backup to completely fail with no workaround, zip and ziparchive both just silently ignore the broken symlink
- `PclZip::create` simply replaces the backup if it exists so complete backups dont work (unit test coming to confirm this), we can't use `PclZip::add` though because that appends duplicate files to an existing zip with no option to skip or replace.


willmot:
Added unit tests to uncover the issues when PclZip is the only backup provider available

pdewouters:
@willmot so, I can filter the backup engines to only include PCLZIP?

willmot:
> @willmot so, I can filter the backup engines to only include PCLZIP for testing?

Yep, see https://github.com/humanmade/backupwordpress/blob/cac1d79dd610af143cb8641559b0c357b22a5e44/tests/class-site-backup/test-class-site-backup.php#L64

The unit tests already do this with since https://github.com/humanmade/backupwordpress/commit/cac1d79dd610af143cb8641559b0c357b22a5e44

willmot:
Worth noting, this PR isn't working as currently the way `Backup` does complete backups isn't compatible with how PclZip works.

willmot:
Moving this out of the 3.6 release as it still requires a non-trivial amount of work to get going.

sambulance:
@willmot Is this still in progress? What needs to happen to get this merged?

willmot:
@sambulance not in progress as I'm unlikely to be able to pick back up. In terms of completing, it needs the two items mentioned in https://github.com/humanmade/backupwordpress/pull/999#issue-136759200 resolving still.

katmoody:
How was it added before? Was it in place of the Unix Zip Command?

PR #1000, opened 2016-02-28
Update to finder 2.8.3

willmot:
We can't update to the 3.x branch as it requires PHP 5.5


pdewouters:
ah, good catch.

composer.json
```diff
@@ -19,7 +19,7 @@
         "issues": "https://github.com/humanmade/backupwordpress/issues"
     },
     "require": {
-        "symfony/finder": "~2.6",
-        "ifsnop/mysqldump-php":"2.*"
+        "symfony/finder": "~2.",
+        "ifsnop/mysqldump-php":"~2."
```
pdewouters: doesn't the star mean any minor version for specified major?

willmot: I was going for the [Next Significant Release syntax](https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators), although looks like I haven't gotten that exactly right. Based on that looks like they should both be ~2.8?

pdewouters:
@willmot I just use whatever is suggested here on the packagist page for the repo  https://packagist.org/packages/ifsnop/mysqldump-php

pdewouters:
ah right, but for Finder you want anything less than the one that requires 5.5?
screenshot 2016-02-29 14 17 53

pdewouters:
so if you have `~2.8.3` it would go up to `2.9` I think

willmot:
> ah right, but for Finder you want anything less than the one that requires 5.5?

Yeah exactly.

> so if you have ~2.8 it would go up to 3.0 I think

I don't think so, the docs say `~1.2 is equivalent to >=1.2 <2.0.0` which is exactly what we want (`~2.8 is equivalent to >=2.8 <3.0.0`)

pdewouters:
yes, that's what I meant, so will you change it to `~2.8`

willmot:
👍

Issue #979, opened 2016-01-27
Incomplete Database


katmoody:
Database backups completed but only showing a partial backup. 

See ticket on wp.org: https://wordpress.org/support/topic/empty-database-dump-after-update?replies=1
See ticket on Intercom: https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1766861518

katmoody:
Interesting coincidence(?) noted in the forums on this same issue affected sites are all using PHP 5.4.39 and sites using other versions are unaffected by this one ... could PHP version impact this in some way? https://wordpress.org/support/topic/empty-database-dump-after-update?replies=9#post-7968190

willmot:
It would be pretty odd for this to be a PHP specific reason, worth checking though as we need to reproduce before we can track down a fix.

willmot:
Do we have access to anyones site / server where this is happening?

katmoody:
I'll ask if the one doing so much troubleshooting would be willing to assist in this way :D He's been the most proactive of the bunch.

RubenMartins:
Hello, This is also happening with me in 5.3.29 php version
Apache/2.2.29 (Unix) mod_ssl/2.2.29 OpenSSL/1.0.1e-fips DAV/2 mod_bwlimited/1.4 PHP/5.3.29

On the version 3.4.2 of the plugin this works fine on the same enviroment, just after update to the latest version of 3.4.3 the database was truncated since then on the database backup and also on the complete backup.

What more information do you need to try to reproduce the error?

I saw that the latest version fix a bug where backups were not downloadable when WordPress installed in a subdirectory could the some changed here?

pdewouters:
@RubenMartins thanks for the report. if you revert to the previous version , does it still happen or does it do a complete backup. there wasn't any change related to db backup between 3.4.2 and 3.4.3

RubenMartins:
@pdewouters Reverting to 3.4.2 also the backup of the database is truncated.
In complete backup or only database backup the database backup file only has this:
```
-- MySQL dump 10.13  Distrib 5.6.28, for Linux (x86_64)
--
-- Host: localhost    Database: XXXXXXXXXXXXXXXXXXX
-- ------------------------------------------------------
-- Server version	5.6.28

/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
[...]

Issue #986, opened 2016-02-05
zip I/O Error


katmoody:
This has been coming up frequently, so listing as a bug but please correct if it means something else? 

```
Zip: zip I/O error: No such file or directory, zip error: Could not create output file (was replacing the original zip file)
```

pdewouters:
just got it on my local install, I'll try and debug, although it's probably related to https://github.com/humanmade/backupwordpress/issues/970

pdewouters:
hmm, weird. I can't reproduce it anymore

katmoody:
In SOME cases, this indicates a larger inability to actually back up their sites; in everyone else it's a nuisance error. 

Clear as mud, right? :D

willmot:
@kat in situations where this error occurs the backup should still work, `zip` would fail and the plugin should then just fallback to `zipArchive`. If that's not happening then if we could get access to someones site where this is reproducible that would help a lot

kat:
I think you have meant @katmoody : in situations where this error occurs the backup should still work, zip would fail and the plugin should then just fallback to zipArchive. If that's not happening then if we could get access to someones site where this is reproducible that would help a lot

willmot:
Sorry @kat! 😊

willmot:
Do we have a way to reproduce this error reliably? There's not enough to go on at the moment.

There's two parts to this that I'm interested in.

1. What is causing that `zip` error? Is it something we have control over?
2. When that error happens why does the backup just fallback to `ZipArchive` and complete successfully? If it does then we probably just shouldn't be surfacing that error. If it doesn't then why not?

willmot:
Dug into this and into point 2. above, found the issue. We weren't correctly checking errors when verifying that the backup had completed successfully. Since 9dbcac6 I'm now almost certain this is going to be fixed in #1001

Issue #970, opened 2016-01-24
Avoid showing users useless backup errors


katmoody:
**Feel free to edit this - I wasn't positive how to categorize it. 

Basically, mysqldump, sh and mysql are all coming up with an error saying they can't be found? 

![screen 20shot 202016-01-23 20at 2021 59 53](https://cloud.githubusercontent.com/assets/5481211/12533686/44bc3da8-c206-11e5-89e4-16b65a32a331.png)

The Intercom ticket is here: https://app.intercom.io/a/apps/7f1l4qyq/inbox/katrina@katsmedia.com/conversations/1767900867

katmoody:
And another ticket now ... I'll add them all under this one if additional tickets involved show up, okay? 

https://app.intercom.io/a/apps/7f1l4qyq/inbox/all/conversations/1783509143

willmot:
👍

pdewouters:
By looking at the screenshots, seems backup are running anyway?

willmot:
Just to note, those errors don't mean that the backup didn't work, just that there were errors in the `mysqldump` command and so the plugin had to fallback to the PHP fallback.

We need to do a better job of only showing relevant errors to users. At the moment users get confused because they see errors like that and think it means their backup didn't work, which isn't true.

katmoody:
Just a note - those are showing despite MySQLDump being installed on the server - it just can't find those commands? With so many seeing that warning, does that mean anything else could be going on aside from it being an ignorable error?

pdewouters:
This is tied to a nexisting issue about showing useful error messages, as we need a smart way to filter what is useful to show to the user. Or can we change the wording as we had before?

> Last backup was successful, generated some warnings that are probably safe to ignore.

willmot:
Couple of points:

- We shouldn't show errors from backup methods that weren't actually used. For example if the plugin tries `zip`, fails and then fallback to `ziparchive` we should just suppress the warnings from `zip`.
- For warnings that do show we should start a whitelist to ignore the ones we no don't actually impact the creation of the backup file.

PR #1001, opened 2016-02-28
Use Process for all our system calls

willmot:
- Replaces all usage of exec
- Removes helper functions and unit tests that dealt with whether exec was available
- Adds Process 2.x to composer
- Process relies on `proc_open` rather than `exec` under the hood so we check for that before using
- Simplifies the errors we bubble up from shell commands, we now only bubble errors when the command reports it failed rather than also bubbling warnings from successful runs.
- Fixes a bug in `verify_backup` that meant we weren't correctly falling back in instances where the process reported an error but the backup file (database or zip) was created (even only partially). Adds unit tests for this.


willmot:
> Simplifies the errors we bubble up from shell commands, we now only bubble errors when the command reports it failed rather than also bubbling warnings from successful runs.

This is will reduce the amount the of useless errors we show people (see https://github.com/humanmade/backupwordpress/issues/970).

willmot:
I'm also interested to see whether this helps with https://github.com/humanmade/backupwordpress/issues/979

@katmoody do we have anyone who's running into the incomplete database issue that could test a beta here?

willmot:
We should consider using `ProcessBuilder` to build our more complex commands: http://symfony.com/doc/current/components/process.html#executing-php-code-in-isolation

willmot:
@pdewouters I think I'm done with this. Would appreciate you testing locally and giving a thorough review/merge.

@katmoody we'll want to beta test this change with a few people who are running into issues. 

I specifically think it will resolve https://github.com/humanmade/backupwordpress/issues/979 & https://github.com/humanmade/backupwordpress/issues/986

pdewouters:
tested on @katmoody 's test site which uses `ZipArchive` and this doesn't conflict and tested locally on an install that has the zip command and backups work ok (db and complete)

willmot:
@pdewouters great! I'm happy to go for merge in that case. We can open a separate issue to cover implementing `ProcessBuilder`.

Once we have everything merged into master we can do a 3.5 beta and see if @katmoody can test with some of the folks who're running into the issues this intends to fix.

pdewouters:
@willmot sorry, deleted previous message, as I realized I hadn't ran composer install ,so was missing the Process library

willmot:
Huh, wouldn't that have caused a fatal?

pdewouters:
```
[29-Feb-2016 11:55:20 UTC] PHP Fatal error:  Uncaught exception 'Symfony\Component\Process\Exception\ProcessTimedOutException' with message 'The process "cd '/srv/www/wordpress-default' && zip -rq '/srv/www/wordpress-default/wp-content/backupwordpress-cbaa44a22d-backups/local-wordpress-dev-1445593823-complete-2016-02-29-11-54-16.zip' ./ -x '*.svn*' -x '*_svn*' -x '*CVS*' -x '*_darcs*' -x '*.arch-params*' -x '*.monotone*' -x '*.bzr*' -x '*.git*' -x '*.hg*' -x '*backwpup-**' -x '*updraft*' -x '*wp-snapshots*' -x '*backupbuddy_backups*' -x '*pb_backupbuddy*' -x '*backup-db*' -x '*Envato-backups*' -x '*managewp*' -x '*backupwordpress-*-backups*' -x '*wp-content/backupwordpress-cbaa44a22d-backups*'" exceeded the timeout of 60 seconds.' in /srv/www/wordpress-default/wp-content/plugins/backupwordpress/vendor/symfony/process/Process.php:1192
Stack trace:
#0 /srv/www/wordpress-default/wp-content/plugins/backupwordpress/vendor/symfony/process/Process.php(364): Symfony\Component\Process\Process->checkTimeout()
#1 /srv/www/wordpress-default/wp-content/plugins/backu in /srv/www/wordpress-default/wp-content/plugins/backupwordpress/vendor/symfony/process/Process.php on line 1192
```

willmot:
@pdewouters made some additional minor improvements:

- No need to increase the timeout for the `mysql --execute=quit` command as that should be quick. If for some reason the cli database connection does take 60 seconds to connect we probably do want to treat that as an error and fallback to PHP anyway.
- Catches `\Exception` instead of `ProcessTimedOutException` in `Zip_File_Backup_Engine` which brings it into line with `Mysqldump_Database_Backup_Engine`.
- Avoids a possible warning in the `test_file_with_strange_characters` unit test by ensuring that the backup file actually exists before attempting to access it's contents.

FYI, for testing purposes you can change the timeout to something like `0.001` to force a timeout Exception.

willmot:
@pdewouters looks like https://github.com/humanmade/backupwordpress/commit/3adf43457cf65becdcad545f3ecc446caacaac13#diff-58e989755f27612a26203553671d2897 was a bad merge, getting a fatal error when trying to load the plugin admin page:

```
Fatal error: Call to undefined method HM\BackUpWordPress\Backup_Utilities::is_exec_available()
```

That was fixed in https://github.com/humanmade/backupwordpress/commit/f33cb44da3e62db90db80f914045dd0766249f1f but looks like it was undone.

Re-fixed in 695c46b

willmot:
- [ ] I need to test this on Windows.

pdewouters:
yeah I did a bad merge, then fixed it. sorry if it caused you an issue

willmot:
> yeah I did a bad merge, then fixed it. sorry if it caused you an issue

No worries, was just noting 👍

willmot:
Unit tests run without issue on Windows, I also manually ran through and smoke tested creating backups etc. I think this is ready for merge 👍

PR #1003, opened 2016-02-29
Only call getPerms on readable files to ensure we avoid a fatal error

willmot:
Bunch of coding standards fixes

### DIFF
--- a/admin/schedule-form-excludes.php
+++ b/admin/schedule-form-excludes.php
@@ -57,7 +57,7 @@ $user_excludes = $excludes->get_user_excludes(); ?>
 
 						<a href="<?php echo admin_action_url( 'remove_exclude_rule', array(
 							'hmbkp_remove_exclude' => $exclude,
-							'hmbkp_schedule_id'    => $schedule->get_id()
+							'hmbkp_schedule_id'    => $schedule->get_id(),
 						) ); ?>" class="delete-action"><?php _e( 'Stop excluding', 'backupwordpress' ); ?></a>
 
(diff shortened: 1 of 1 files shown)


admin/schedule-form-excludes.php
```diff
@@ -303,24 +291,21 @@ class="dashicons dashicons-update"
 					</td>
 
 					<td>
-						<?php echo esc_html( substr( sprintf( '%o', $file->getPerms() ), - 4 ) ); ?>
+						<?php if ( ! $is_unreadable ) {
```
willmot: This is the line which matters, we needed to check if the file was readable before calling `getPerms`

admin/schedule-form-excludes.php
```diff
@@ -103,102 +102,98 @@
...
+				</th>
 
-					$root = new \SplFileInfo( Path::get_root() );
+				<td class="column-filesize">
 
-					$size = $site_size->filesize( $root );
+					<?php if ( Site_Size::is_site_size_being_calculated() ) { ?>
 
-					if ( false !== $size ) {
+						<span class="spinner is-active"></span>
```
pdewouters: missing opening bracket?

admin/schedule-form-excludes.php
pdewouters: where is the var defined?

willmot: Just here: https://github.com/humanmade/backupwordpress/blob/fatal-get-perms/admin/schedule-form-excludes.php#L215

PR #1008, opened 2016-02-29
Switch to storing site size data in a file

willmot:
Instead of storing an array of all files and their associated filesizes in the DB in a transient we now store it in a hidden `.files` file.

The array of data is `json_encoded` into a string and then compressed with `gzcompress` this helps keep filesizes down.

Adds a unit test to ensure the file is only cached for a week.

### DIFF
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -536,7 +536,7 @@ function heartbeat_received( $response, $data ) {
 		if ( ! empty( $data['hmbkp_client_request'] ) ) {
 
 			// Pass the site size to be displayed when it's ready.
-			if ( Site_Size::is_site_size_cached() ) {
+			if ( Site_Size->is_site_size_cached() ) {
 
 				$site_size = new Site_Size( $schedule->get_type(),  $schedule->get_excludes() );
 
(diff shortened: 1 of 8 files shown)


willmot:
On my fairly vanilla WordPress install the `.files` file ends up being just under `75 KB`. It would be useful to test this on a large site.

pdewouters:
do we want to maybe handle write/read errors?

pdewouters:
`[01-Mar-2016 09:29:43 UTC] PHP Strict standards:  Non-static method HM\BackUpWordPress\Site_Size::is_site_size_cached() should not be called statically in /srv/www/wordpress-default/wp-content/plugins/backupwordpress/admin/actions.php on line 539`

willmot:
Fixed those issues, thanks for catching @paulgibbs 👍

I'm also going to test this on Windows now.

willmot:
- [x] Tested on Windows

willmot:
> do we want to maybe handle write/read errors?

pdewouters:
do we want to delete the old transient on plugin upgrade?

willmot:
> do we want to delete the old transient on plugin upgrade?

Done

pdewouters:
weird, on this large site I have, the site size calculation doesn't seem to finish.

PR #1010, opened 2016-03-01
Improve how filesizes are handled in the Your Site list

willmot:
- The list of files and folders is now ordered by the included filesize, that is the total filesize minus any excludes
- We now display the included filesize for directories in the format "Included size of total size".
- All cells in the table are now wrapped in `<code>` tags
- Unreadable files now show `-` for size and permissions to indicate that we don't actually have the data

Here's how that looks:

screen shot 2016-03-01 at 00 01 50

### DIFF
--- a/admin/schedule-form-excludes.php
+++ b/admin/schedule-form-excludes.php
@@ -17,56 +17,56 @@ $user_excludes = $excludes->get_user_excludes(); ?>
 
 		<tbody>
 
-		<?php foreach ( $user_excludes as $key => $exclude ) :
+			<?php foreach ( $user_excludes as $key => $exclude ) :
 
-			$exclude_path = new \SplFileInfo( trailingslashit( Path::get_root() ) . ltrim( str_ireplace( Path::get_root(), '', $exclude ), '/' ) ); ?>
+				$exclude_path = new \SplFileInfo( trailingslashit( Path::get_root() ) . ltrim( str_ireplace( Path::get_root(), '', $exclude ), '/' ) ); ?>
 
-			<tr>
+				<tr>
 
-				<th scope="row">
+					<th scope="row">
 
-					<?php if ( $exclude_path->isFile() ) { ?>
+						<?php if ( $exclude_path->isFile() ) { ?>
 
-						<div class="dashicons dashicons-media-default"></div>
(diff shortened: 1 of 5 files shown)


willmot:
Verified this on Windows too

pdewouters:
I'll review after https://github.com/humanmade/backupwordpress/pull/1008 is merged

admin/schedule-form-excludes.php
```diff
@@ -154,40 +155,32 @@
...
-							if ( ! $size ) {
-								$size = '0 B';
-							} ?>
+						$excluded_size = $excluded_site_size->filesize( $root ); ?>
 
 							<code>
 
-								<?php echo esc_html( $size ); ?>
+								<?php
+								echo ( preg_replace( '/[0-9]+/', '', size_format( $size ) ) === preg_replace( '/[0-9]+/', '', size_format( $excluded_size ) ) ) ? esc_html( (int) size_format( $excluded_size ) ) : esc_html( size_format( $excluded_size ) );
```
pdewouters: I find it difficult to follow what's going on here. Also, should it maybe be a `sprintf` so we can make it translatable?
Could add a translators comment to explain what the `%1$s` are

willmot: @pdewouters good point re. translation, I'll add that and try to improve the code readability

willmot:
@pdewouters I extracted some of the logic out into a function and made those strings translatable, see https://github.com/humanmade/backupwordpress/commit/cb625ad815c2edf4d94c5e847a1cdd6fda33450c

willmot:
Given `size_format` seems to fairly aggressively round we could look at showing more decimals, however we'd probably want to be smart about it as showing decimals when they're not really needed doesn't look great:

screen shot 2016-03-03 at 11 50 29

- [ ] Don't show decimals for `.00`
- [ ] Don't show decimals for Bytes as that makes no sense
- [ ] Consider doing some small amount of rounding and only showing decimals when it's in the `> .25` and `< .75` range.

willmot:
@pdewouters I think we should merge this and tackle https://github.com/humanmade/backupwordpress/pull/1010#issuecomment-191724396 as a separate PR

Issue #854, opened 2015-07-29
Do we need no-create-db flag in mysqldump


pdewouters:
From the mySql docs: 
> This option suppresses the CREATE DATABASE statements that are otherwise included in the output if the --databases or --all-databases option is given.

We don't use those flags so I think we can remove it?

willmot:
Where are we using it?

If there are `CREATE DATABASE` statements in the dump file then won't they cause an error if you try to import the dump into an existing (but empty) database?

pdewouters:
here: https://github.com/humanmade/backupwordpress/blob/master/classes/class-backup.php#L838

The way I understand that sentence is IF you use `--databases` or `--all-databases` then you should use `no-create-db` to suppress `CREATE DATABASE` statements

willmot:
Ah I see what you mean, we should check if that's the case, if so, seems like we could safely remove.

PR #1011, opened 2016-03-01
Remove the no-create-db flag

willmot:
it's only useful when paired with the `--database` or `--all-databases flags`


willmot:
Fixes https://github.com/humanmade/backupwordpress/issues/854

PR #1012, opened 2016-03-01
Normalize paths before working with them

willmot:
Fixes root path issues on Windows


willmot:
See https://github.com/humanmade/backupwordpress/issues/976, this likely fixes the issue that was re-opened for, although we don't have full details so can't be totally sure.

pdewouters:
good catch, I thought I had done that.

Issue #908, opened 2015-10-09
Cancel backups that are killed. 


owaincuvelier:
It's impossible to detect if the process has been killed, but we can safely assume backups won't take longer then 45 minutes or so. 

If the backup still seems to be running after 45 minutes to an hour, cancel it.

willmot:
My names Tom Willmot and I endorse this message

pdewouters:
thank you sir

owaincuvelier:
:+1:

owaincuvelier:
We have some users now waiting for this to be released. Their aware and understanding of backpus crashing, but having to manually click the cancel button from time to time is inconvenient. Backups are quite reliable, except for the occasional win. I think this would be a big win so I hope we can release this soon!

katmoody:
FYI - user suggestion that in case of failed backup, an email might be sent to the user to let them know. As otherwise they are assuming all is well. :-)

willmot:
@katmoody we have an existing ticket covering sending emails to users when the backup fails

katmoody:
We are still having requests for this one, just letting you know ;-)

willmot:
> It's impossible to detect if the process has been killed

Actually it's not, see #1025. But, we could still look at timing backups out as a further catch for any issues that somehow slip through #1025.

Issue #763, opened 2015-03-13
Handle PHP fatal errors that occur during backup


pdewouters:
Now that we are not waiting for the ajax response, we should implement a different way to catch PHP fatal errors and handle them and display a notice to the user.

willmot:
Closing in favour of #1025

Issue #767, opened 2015-03-18
notify user if scheduled backup went wrong


pdewouters:
do scheduled backups fail silently? the problem is an error can go unnoticed for weeks until the user tries to use a backup to restore their site only to notice that something is missing.
we should try and report any errors or check the zip and make sure it's complete

willmot:
Yeah they basically do, problem is if cron aren't working then the schedule code never runs.

We could write some code which checks the previous scheduled backups have run as expected that runs when the user visits the admin, but we couldn't email them at the time.

willmot:
> we should try and report any errors or check the zip and make sure it's complete

We do catch errors and if the user has selected to have the backup emailed to them then we also email them backup errors. Perhaps we should just email everyone backup errors?

pdewouters:
I'm down with that

pdewouters:
I think we should ask permission before sending emails? I don't like adding new settings but it should be some kind of opt-in

willmot:
I'm not so sure. If you're using a service and it fails you would expect to be notified. If you don't want to be notified of errors then you mustn't care that backups are failing, in-which case you can just opt-out by removing the plugin.

I think the fact that without notifications there are people who have backups failing but don't know about it is worse than the fact that people who don't want email would start getting them. If they don't want them they can remove the plugin.

Said differently I think these kinds of emails would be considered critical service emails and thus wouldn't need to be opt-in.

pdewouters:
We should implement a notification service which allows different types of notifs such as email, slack message, push notif, sms,...

willmot:
They would make good extensions actually, nice idea!

Issue #752, opened 2015-03-06
Backup files in batches to avoid timeouts


pdewouters:
Right now when we start a backup, the process can generally take more than 30 seconds which is usually the PHP timeout limit.
We could use the file manifest method to start a backup and time it. Each batch should not run for more than 10 - 20 seconds to be safe. We keep note of where we are so we also can resume backup if it fails

We would create a subfolder per backup with the archive parts in it. Instead of manually creating the multiple zip files, let's use zip's split size functionality.
` zip -s 670m -r foo bar`

Doesn't seem that you can do delta backups at the same time though. I do think that this method would be more beneficial if we can avoid timeouts.

>  Use  -s  to set the split size and create a split archive.  The size is
       given as a number followed optionally by one of k (kB), m (MB), g (GB),
       or  t (TB) (the default is m).  The -sp option can be used to pause zip
       between splits to allow changing removable media, for example, but read
       the descriptions and warnings for both -s and -sp below.

       Though  zip does not update split archives, zip provides the new option
       -O (--output-file or --out) to allow split archives to be  updated  and
       saved in a new archive.  For example,

              zip inarchive.zip foo.c bar.c --out outarchive.zip

willmot:
If we have access to `zip` then we won't be hitting PHP timeouts as time taken performing system commands isn't counted towards the PHP execution time.

Only `PclZip` and `ZipArchive` should be hitting PHP timeouts.

Even if we are hitting PHP timeouts with `zip` it would be caused by something other than the actual `zip` process and so splitting into delta wouldn't help.

pdewouters:
ah, doesn't running `exec` or `shell_exec` pause execution until it gets a response?

pdewouters:
Just did a little digging and yes, you're right.It doesn't hit a PHP timeout error, but it can hit the server gateway timeout error.
Also, memory usage can be intensive

willmot:
Using `zip`s `-s` flag to create a split archive wouldn't help with either the time it takes nor the memory usage as it still does the whole thing in a single process, it's just that it creates multiple files instead of a single file.

[...]

PR #1025, opened 2016-03-09
Utilise file locking so we can track crashed processes

willmot:
We are currently [unable to tell whether the background PHP process that the backup is running in has crashed](https://github.com/humanmade/backupwordpress/issues/908) or is still running.  We create a `.backup_running` file to track that a backup is in progress, this file is then deleted when the backup completes. The problem is that if the backup process fatal errors, crashes or is killed for some reason that file is not deleted and thus the backup looks like it's still running.

This PR solves this issue with [file locking], which is supported in PHP with the [`flock`](http://php.net/manual/en/function.flock.php) function. Once set, a file lock ensures that the file cannot then be written to by another thread until the lock is removed by the first thread. The kicker for us, file locks are automatically cleaned up when a PHP process ends, no matter how it's ended. We can use that fact to detect crashed backups.

If the `.backup_running` file exists and is locked then the backup must still be running and crucially if the `.backup_running` file exists but the isn't locked then the backup process must have started and then ended before it could properly complete. Huzzah!

We're using [`symfony/LockHandler`](http://symfony.com/doc/current/components/filesystem/lock_handler.html) to make working with file locks easier.


willmot:
This still needs more testing and ideally some robust unit tests before it's merged. Unfortunately unit testing is complex as it requires starting backups in multiple threads and then killing them. Work continues.

willmot:
This would fix https://github.com/humanmade/backupwordpress/issues/908 and will help with https://github.com/humanmade/backupwordpress/issues/763, https://github.com/humanmade/backupwordpress/issues/762 & https://github.com/humanmade/backupwordpress/issues/752

willmot:
Added in detection of crashed backups, we now cleanup properly and show a message:

screen shot 2016-03-12 at 01 10 01

#### Still Todo

- [ ] Mixup the unit tests
- [ ] Better error message
- [ ] Persistent notices aren't clearing up properly
- [ ] Test on Windows
- [ ] Check on a system which doesn't support file locking, we shouldn't default to `has_crashed`

willmot:
https://github.com/humanmade/backupwordpress/commit/abb7e410a16bbf0fb25de477ec79896fedbf5b48 fixes https://github.com/humanmade/backupwordpress/issues/763

willmot:
@pdewouters we should get a beta of this release going so Kat can test with people who are running into the issues this should solve.

katmoody:
Consider adding one further item above @willmot and/or @pdewouters -- sending an email when this comes up? Or will that automatically send as per the email settings? I get a lot of notes from folks who never got a warning anything was wrong on their backups and don't check their admin regularly, so I'm thinking the email warning would be an important aspect.

willmot:
@katmoody agreed that would be a very useful enhancement, we actually have that covered in #767

willmot:
> Fixup the unit tests

I think unit testing with multiple PHP threads is going to be a blocker here. For now the unit tests are `markSkipped`.

willmot:
On systems which don't support file locks there are 2 error conditions that will in practise mean backups will stop working for those users.

1. Locking never works: backups won't be able to start as the status file won't be locked in `Backup_Status::start()`.

2. If the status file is locked when starting the backup but checking the lock fails for some reason then the backup will be (incorrectly) marked as crashed and deleted before it can finish.

willmot:
We also need to look into the consequences of this warning from [php.net](http://php.net/manual/en/function.flock.php)

> On some operating systems flock() is implemented at the process level. When using a multithreaded server API like ISAPI you may not be able to rely on flock() to protect files against other PHP scripts running in parallel threads of the same server instance!

Specifically, what does this mean for locking files in multi-server setups?

willmot:
> Specifically, what does this mean for locking files in multi-server setups?

Here's my understanding:

1. `flock` file locking is process-based. That means the locks are only respected by the current PHP process. Multiple threads from the same process work correctly.
2. In multi-server environments it'll depend on whether the shared filesystem supports `flock` or not. 
    - NFS support `flock`.
    - S3 Uploads won't.

- [ ] We should provide a Constant to bypass it. That way, users who's FS doesn't support locking have a way forward.

willmot:
I think this one is ready to go!

We'll want to do a beta before it's released though so let's leave it in it's own branch for now.

katmoody:
@willmot - do we want to test this on folks whose sites aren't consistently running backups because of the spinning wheel of doom? Is it ready for beta testing in that case?

willmot:
@katmoody yep exactly, once Dasha has reviewed I'll create a beta release zip that we can share with people.

dashaluna:
@willmot I've looked at the code, however, my server knowledge isn't great. So, I don't understand what's going on :( Would you mind assigning the review of this to someone else? Apologies.

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
 
+	/**
+	 * The filename for the backup file we are the tracking status of.
+	 *
+	 * @var string
+	 */
 	private $filename = '';
 
+	/**
+	 * [$lock_handler description]
```
rmccue: This needs an actual description.

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
+	 *
+	 * @var LockHandler
+	 */
+	private $lock_handler = '';
+
+	private $callback;
```
rmccue: Needs phpDoc

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
+	 */
 	public function is_started() {
 		return (bool) file_exists( $this->get_status_filepath() );
 	}
 
-	public function finish() {
-		// Delete the backup running file
-		if ( file_exists( $this->get_status_filepath() ) ) {
-			unlink( $this->get_status_filepath() );
+	public function is_running() {
```
rmccue: phpDoc

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
+
+			// If we're in the same thread then we know we must be running if the running file exists.
+			if ( is_a( $this->lock_handler, 'LockHandler' ) ) {
+				return $this->is_started();
+			}
+
+			$lock_handler = new LockHandler( basename( $this->get_status_filepath() ), Path::get_path() );
+
+			return ! $lock_handler->lock();
+		}
```
rmccue: This conditional should be flipped, and returned early instead.

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
+	 */
+	public function catch_fatals() {
+
+		$error = error_get_last();
+
+		if ( empty( $error ) ) {
+			return;
+		}
+
+		if ( ! isset( $error['type'] ) || ! defined( 'E_ERROR' ) || E_ERROR !== $error['type'] ) {
```
rmccue: `E_ERROR` is always defined; should also catch `E_USER_ERROR` here.

classes/backup/class-backup-status.php
```diff
@@ -2,44 +2,209 @@
...
+
+		if ( $this->is_started() ) {
+			$status = json_decode( file_get_contents( $this->get_status_filepath() ) );
+
+			if ( ! empty( $status->filename ) ) {
+				$this->filename = $status->filename;
+			}
+		}
+
+		return $this->filename;
```
rmccue: Seems a bit weird to always return this even if we haven't actually gotten a filename from the file

classes/backup/class-backup-status.php
```diff
@@ -105,15 +272,18 @@ public function get_start_time() {
...
-	 * Get the path to the backup running file that stores the running backup status
+	 * Get the path to the backup running file that stores the running backup status.
 	 *
 	 * @return string
 	 */
 	public function get_status_filepath() {
 		return Path::get_path() . '/.backup-' . $this->id . '-running';
 	}
+
+	public function set_status_callback( $callback ) {
```
rmccue: phpDoc

willmot:
@sambulance  @shadyvb I've struggled to find time to fixup Ryan's comments above. If either of you want to take a bash it would be great to get this in.

Issue #1031, opened 2016-03-13
Improve code quality


willmot:
We have a variety of ways of measuring the code quality of our plugin:

- The [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards).
- [Scrutiniser](https://scrutinizer-ci.com/g/humanmade/backupwordpress/).

Refactoring our code with the sole purpose of reducing the issues reported by these tools is risky, however over the long term we should be working with towards improving our overall code quality.

This ticket will likely never be closed as there's always improvements to be made. However it's a useful place to collate the various issues and track conversation and progress.

willmot:
Since #1032 we're down to only two sets of WPCS issues:

- [ ] We're silencing numerous function calls, for example wherever we call filesystem functions directly (`fopen` etc) or when calling functions which might be disabled (`ini_set` for example). This is inline with what WordPress Core does and makes sense in our case as we genuinely have no way of knowing whether those function calls will throw a PHP error and if they do we really do want to just silently fail and fallback.
- [ ] There are some indentation errors which appear to be bugs in WPCS, for example on my local setup, WPCS throws an error at this line: https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-form-excludes.php#L258 reporting `Line indented incorrectly; expected 1 tabs, found 7.` which is clearly incorrect.
<img width="831" alt="screen shot 2016-03-13 at 14 04 22" src="https://cloud.githubusercontent.com/assets/308507/13729092/82445012-e924-11e5-974c-5a762f69dde2.png">

willmot:
For our silenced function calls, I'm we should potentially audit our use to make sure we're not silencing when we shouldn't be (I don't think we need to silence `ini_get` for example as Core doesn't). We can then wrap them in a `// @codingStandardsIgnoreStart` block.

willmot:
We have a lot of noise in our [scrutinizer issue report](https://scrutinizer-ci.com/g/humanmade/backupwordpress/issues/master) which makes finding and fixing actual issues a paid. I've been through and added a bunch of [issue filters](https://scrutinizer-ci.com/g/humanmade/backupwordpress/settings/issue-filters?page=1) to cover things we're not interested in changing. Documenting it here so we have a record.

PR #1032, opened 2016-03-13
WPCS fixes

willmot:
Fixes nearly all [WPCS](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) errors across all the PHP files in the plugin.

We're using the `WordPress-Core` set of sniffs.

See https://github.com/humanmade/backupwordpress/issues/1031


willmot:
@pdewouters I know code standards fixes for the sake of code standards fixes are generally a bad idea, however we're so close to almost completely fixing all the smells that I thought it worth doing the last few as a single PR.

willmot:
I'm going to leave the branch around, we can continue pushing WPSC fixes to it, opening PR's and re-merging.

PR #1034, opened 2016-03-14
Add unit tests for the Notices class and fix bugs

willmot:
This PR fixes up a bunch of issues in the Notices class and adds full unit test coverage. Issues resolved:

 * `(array) get_option` returned an array with a single empty element rather than an empty array.
 * Setting a persistent notice duplicated it as it was then in `$this->notices` and the database.
 * Calling `get_notices` with a context which didn't exist returned all notices rather than none.
 * Calling `set_notices` multiple times with same context overwrote all but the last messages.
 * It was possible to set broken messages with empty context or message.
 * `update_option` was needlessly called when setting a message that had already previously been set.

It also completes the docblocks and brings the Singleton pattern inline with Path.

- [x] Persistent notices aren't dismissible.

### DIFF
--- a/classes/class-notices.php
+++ b/classes/class-notices.php
@@ -23,7 +23,7 @@ class Notices {
 	/**
 	 * Contains the instantiated Notices instance
 	 *
-	 * @var Path $this->get_instance
+	 * @var Notices $this->get_instance
 	 */
 	private static $instance;
 
--- a/classes/class-notices.php
+++ b/classes/class-notices.php
@@ -115,9 +115,9 @@ class Notices {
 
 		$notices = $this->get_all_notices();
 
-		if ( $notices ) {
+		if ( ! empty( $notices ) ) {
 
-			if ( $context ) {
+			if ( ! empty( $context ) ) {
 				return isset( $notices[ $context ] ) ? $notices[ $context ] : array();
 			}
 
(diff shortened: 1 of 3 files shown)


willmot:
Would be good to catch the Scrutiniser issue before merging: https://scrutinizer-ci.com/g/humanmade/backupwordpress/inspections/fb6bf204-3ded-42f2-8c0e-21b8cf8a3b01/issues/files/classes/class-notices.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

classes/class-notices.php
```diff
@@ -76,19 +128,33 @@ public function get_notices( $context = '' ) {
...
+	 *
+	 * @return array The array of notices
 	 */
-	public function clear_all_notices() {
+	private function get_persistent_notices() {
+		$notices = get_option( 'hmbkp_notices' );
+		if ( ! empty( $notices ) ) {
+			return $notices;
+		}
+		return array();
```
pdewouters: How about `return ! empty( $notices ) ? $notices : array();`

willmot: Yeah that would be neater 👍 doing now.

classes/class-notices.php
```diff
@@ -65,8 +117,8 @@ public function get_notices( $context = '' ) {
 
 		if ( $notices ) {
```
pdewouters: Should it be `if ( ! empty( $notices ) )` ?

pdewouters:
is it good to review now?

willmot:
@pdewouters yep!

Issue #1040, opened 2016-03-18
Stop using `mysql` to check cli database access


willmot:
We're running into issues where the `mysql` isn't available but `mysqldump` is, in those instances we cause users to fallback to `PDO` when in reality it would be fine to use `mysqldump`.

Let's just remove that check and let mysqldump run, it will fallback anyway.

katmoody:
Is this error an indication of this issue @willmot? 

```Mysqldump: mysqldump: Got error: 145: Table './ArabiaHolidaysCom/wp_wfLeechers' is marked as crashed and should be repaired when using LOCK TABLES

IMysqldump: SQLSTATE[HY000]: General error: 145 Table './ArabiaHolidaysCom/wp_wfLeechers' is marked as crashed and should be repaired```

If not, I'll open a new ticket - but the backup works despite this error, making me think that the error is coming up on the MySQLDump but the fallback is still working. (also - should I tell Wordfence about that table crashing - "wp_wfLeechers" is one of theirs I think - it's the third time that error has come up on one of the sites contacting us?)

willmot:
@katmoody nope different issue, they have crashed tables and need to repair them. See https://github.com/humanmade/backupwordpress/issues/836

PR #1055, opened 2016-04-03
Stop testing the mysql connection directly

willmot:
The error will be captured in mysqldump anyway and using mysql is an imperfect test which opens the door to false negatives.


willmot:
Fixes https://github.com/humanmade/backupwordpress/issues/1040

pdewouters:
needs syncing with master branch

willmot:
@dashaluna would be great if you could re-merge master here so we can get this in.

PR #1071, opened 2016-06-06
Stop running filesizes through size_format twice.

### LINKED ISSUE #1058: Incorrect File Sizes Recorded under Exclusions
Pretty sure this might be universal - have seen on every site I've looked at running BWP. 

When you go under exclusions, the files sizes are listed incorrectly. So if all the files together are 600 MB, and the uploads folder is actually 300 MB in size, it just shows 300 B in size. 

tools php 1

willmot:
Fixes #1058 issues with excludes filesizes showing incorrectly.

### DIFF
--- a/admin/schedule-form-excludes.php
+++ b/admin/schedule-form-excludes.php
@@ -253,8 +253,8 @@ $user_excludes = $excludes->get_user_excludes(); ?>
 							$size = $site_size->filesize( $file );
 
 							if ( false !== $size ) :
-								$size = size_format( $size ) ?: '0 B';
-								$excluded_size = size_format( $excluded_site_size->filesize( $file ) ) ?: '0'; ?>
+								$size = $size ?: '0';
+								$excluded_size = $excluded_site_size->filesize( $file ) ?: '0'; ?>
 
 								<code>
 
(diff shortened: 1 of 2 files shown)


dashaluna:
@willmot related WP bug? https://github.com/WordPress/WordPress/blob/master/wp-includes/functions.php#L262

admin/schedule-form-excludes.php
```diff
@@ -8,10 +8,10 @@
 <div class="hmbkp-exclude-settings">
 
 	<h3>
-		<?php _e( 'Currently Excluded', 'backupwordpress' ); ?>
+		<?php esc_html_e( 'Currently Excluded', 'backupwordpress' ); ?>
 	</h3>
 
-	<p><?php _e( 'We automatically detect and ignore common <abbr title="Version Control Systems">VCS</abbr> folders and other backup plugin folders.', 'backupwordpress' ); ?></p>
+	<p><?php esc_html_e( 'We automatically detect and ignore common Version Control Systems folders and other backup plugin folders.', 'backupwordpress' ); ?></p>
```
dashaluna: I've got rid of `abbr` here, I don' think there is much for the benefit of having it - I'd prefer "don't make me think" approach when it's spelt straight away. Also makes escaping easier ;)

dashaluna:
As I've mentioned I will be fixing code that I work on to adhere to WP Coding Standards.

admin/schedule-form-excludes.php
willmot: 👍 agreed

willmot:
> As I've mentioned I will be fixing code that I work on to adhere to WP Coding Standards.

👌

dashaluna:
@pdewouters ready for review & merge

willmot:
@dashaluna thanks for fixing up the unescaped translations as well 👍

PR #1077, opened 2016-07-01
Admin performance improvements

willmot:
Some attempts to reduce the impact BackUpWordPress has on admin performance.

### DIFF
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -411,6 +411,7 @@ function add_exclude_rule() {
 	$schedule->set_excludes( $exclude_rule, true );
 
 	$schedule->save();
+	delete_transient( 'hmbkp_root_size' );
 
 	wp_safe_redirect( wp_get_referer(), '303' );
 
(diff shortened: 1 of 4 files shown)


willmot:
@dashaluna @pdewouters could one of you review/merge and then we can release.

pdewouters:
@willmot did you do any benchmarks?

willmot:
Yep, locally for me this cut general page load performance in half and dropped the backups page itself about 8x.

I used query monitor to track which queries were getting called on those pages and then just transient cached around each of them.

pdewouters:
@willmot `set_transient` 3d argument should be `Time until expiration in seconds from now`, not a unix timestamp

willmot:
Ah good catch 👍

dashaluna:
@willmot all fixed, I haven't run unit tests, as I'm still working on setting it up. Do you want to give it a run on unit tests and merge if they are ok?

Issue #1072, opened 2016-06-07
Rethink display of excluded folder/files from `Your Site` table


dashaluna:
How excluded folders/files and their corresponding size should be displayed in the `Your Site` table. At the moment it's confusing as it lists excluded resources in the `Your Site` table with their size and a mark that it's been excluded. In this case what will be intuitive to displaying excluded file sizes?

Discussion originated in issue #1058 and move here into a separate issue.

![manage backups wordpress org wordpress](https://cloud.githubusercontent.com/assets/2185972/15856806/e7a919fa-2caf-11e6-8e41-c98f71849ed6.png)

dashaluna:
Other questions / brainstorming:

1. Why excluded folders displaying at the bottom of the list even though originally they where at the top? Excluded files don't seem to change position in the list.
2. I wonder if more visual aid will be helpful to highlight the excluded files - different row background or font colour?
3. For all excluded folders and files, we could display `0 of **` where `**` is the actual size of a folder/file.
4. We could remove excluded folders/files from the `Your Site` table. However, then a user won't be able to navigate those folders - as it's possible at the moment. However, excluded folders are only traversable, but you can't `stop excluding` from there.

willmot:
I think we can definitely improve this.

> Why excluded folders displaying at the bottom of the list even though originally they where at the top? Excluded files don't seem to change position in the list.

That's a side effect of them showing as `0` size as the list is ordered by included filesize. I agree it's super confusing that folders show as `0` whereas files show their original size. I agree that just showing `0 of xx` for files as well would be the simplest fix here.

> I wonder if more visual aid will be helpful to highlight the excluded files - different row background or font colour?

I'd be up for that, having them all at the bottom will help deemphasize them but a different background colour or something could also help. Feel free to have a play around and drop some example screenshots here.

> We could remove excluded folders/files from the Your Site table. However, then a user won't be able to navigate those folders - as it's possible at the moment. However, excluded folders are only traversable, but you can't stop excluding from there.

I think removing them entirely could be confusing as someone might use that table to find a specific file or folder to check whether it's actually excluded or not.

I like the idea of being able to stop excluding from that list but there is a complication there if you were to try to stop excluding a single file / subfolder that is itself inside a parent which is excluded.  For example, if you exclude `wp-content`, then navigate inside it, you couldn't easily `stop excluding` just `wp-content/uploads`.

willmot:
[...]

PR #1078, opened 2016-07-06
1072: Fix excluded files and directory included sizes

dashaluna:
Issue: #1072


admin/schedule-form-excludes.php
```diff
@@ -298,12 +298,15 @@
 							$size = $site_size->filesize( $file );
 
 							if ( false !== $size ) :
-								$size          = $size ?: '0';
-								$excluded_size = $excluded_site_size->filesize( $file ) ?: '0'; ?>
+								$size          = $size ?: 0;
+								$excluded_size = $excluded_site_size->filesize( $file ) ?: 0;
```
dashaluna: @willmot currently excluded files display size `3 of 3 KB` for example, due to this line. If you follow the function call `filesize()` you can see that for files it returns their size regardless if they are excluded or not, ref: https://github.com/humanmade/backupwordpress/blob/master/classes/class-site-size.php#L184

However, the function for calculating `directory_filesize()` in the same class is aware of excluded files - so that directory size is calculated without excluded files sizes.

Question: should we improve the `filesize()` in the `Site_Size` class to make it aware of excluded files? I.e. if the file is excluded then return `0`, otherwise the file size. I think that the class it the right place for it, as calculating sizes is its purpose. However, I wanted to double check.

dashaluna: Actually, I think it should be this function returning `0` for excluded files, so that down the line they get sorted to the bottom of the list of sizes. I'll double check if this is used anywhere else.

willmot: @dashaluna that sounds exactly right, let's just return `0` in `Site_Size::filesize` if the file is excluded.

dashaluna: @willmot how does this look? Are you happy with `0 B of x KB` or shall I look into removing format after `0`?
screen shot 2016-07-07 at 14 20 10

willmot: Much better!

I think we could remove the `B` but not hugely important, we could do that separately in order to ship this more quickly if you want.

dashaluna:
@willmot this would need a unit test run before merging (I'm still working on setting it up).

willmot:
@dashaluna the unit tests run automatically via Travis, that's what the "All checks have passed" thing above means. 
screen shot 2016-07-07 at 14 46 01
You can click that to expand it:
screen shot 2016-07-07 at 14 46 17
And then click details to jump over to Travis and look at the actual test results:
screen shot 2016-07-07 at 14 46 52
You can then take a look at a specific run: https://travis-ci.org/humanmade/backupwordpress/jobs/143045894

classes/class-site-size.php
```diff
@@ -163,32 +163,41 @@ public function recursive_filesize_scanner() {
...
 			return 0;
 		}
 
-		// If it's a file then just pass back the filesize
+		// If it's a file then return its filesize or 0 if it's excluded.
 		if ( $file->isFile() ) {
-			return $file->getSize();
+
+			if (
+				$this->excludes
```
willmot: I _think_ this will only catch exact exclude matches but not files that are excluded because one of it's parent folders was excluded. Instead you'll need to do a `preg_match` ala https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-form-excludes.php#L236

dashaluna: You're totally right! Missed that one..

willmot: Yeah good idea 👍

dashaluna:
@pdewouters I'd need some help with writing unit test please :) I've added function signatures with comments inside of what I'd like to test. But not sure how to do that. Could you please have a look at it and let me know what you think?

I'd like to test changed/introduced functionality in this PR.

pdewouters:
@dashaluna sure, I'll take a look on Monday

classes/class-excludes.php
```diff
@@ -195,10 +195,6 @@ function( $exclude ) {
 	 */
 	public function is_file_excluded( \SplFileInfo $file ) {
 
-		if ( ! $file->isFile() ) {
```
dashaluna: @willmot I don't understand this 100%. I thought it only worked for files, not folders.. I've moved the original code logic from https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-form-excludes.php#L233 into a separate function in https://github.com/humanmade/backupwordpress/pull/1078/files#diff-1852fb143db94269ade1ed75930ddc87R196

However, if I have that check the wholly excluded Folders get the "Exclude" action button, instead the correct "Excluded" text note. So I removed this check and all works as before.

I'm just a bit confused as to why.

dashaluna:
@willmot also, could you scan through the unit test please, to make sure they make sense? I've run all of them - they work. But wanted to make sure that they are sound as well and if I haven't missed any cases?

classes/class-excludes.php
willmot: The `$files` array in https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-form-excludes.php#L233 contains both directories and files so the new `is_file_excluded` method will need to handle both too. Perhaps renaming the method to `is_path_excluded` would make more sense.

tests/test-excludes.php
```diff
@@ -261,6 +262,77 @@ public function testWildCard() {
...
+		// Check the file is created and its size is NOT 0.
+		$this->assertContains( $this->root->getPath(), $file->getPath() );
+		$this->assertNotSame( $file->getSize(), 0 );
+
+		// Exclude file directly.
+		$excluded_file = new Excludes( $file_path );
+
+		// Check the file is excluded - true.
+		$this->assertContains( basename( $file->getPathname() ), $excluded_file->get_user_excludes() );
+		$this->assertSame( $excluded_file->is_file_excluded( $file ), true );
```
willmot: There's also `assertTrue`

willmot:
@dashaluna these look great, happy for it to be merged 👍

dashaluna:
@willmot let me clean up it a bit - rename the function and unit tests.

classes/class-excludes.php
dashaluna: @willmot What I got confused about is that `$files` array is getting values from `list_directory_by_total_filesize()` https://github.com/humanmade/backupwordpress/blob/master/admin/schedule-form-excludes.php#L114 which internally uses `Site_Size->filesize( \SplFileInfo $file )` and that takes `\SplFileInfo` param, which according to docs is `...offers a high-level object oriented interface to information for an individual file`.

That's why I got confused as to why does it work for directories? Shall I still rename my newly added function to `is_path_excluded`?

willmot:
@dashaluna I made a couple of additional changes. Most notably, to only show `x of x` if some files in the directory have been excluded (and thus the numbers aren't the same).

dashaluna:
@willmot couple of questions/notes:

1. I can still see `0 of xx MB` for directories that are fully excluded (see screenshot). Actually I'd argue that seeing that information is helpful, so that I know how much more will be added to the overall backup size if I'm to include that folder. What do you think?
screen shot 2016-09-01 at 10 47 30

2. Re `int` and `null` return values in `directory_filesize()`. Do we need to add unit test for that?

willmot:
> 1. I can still see `0 of xx MB` for directories that are fully excluded

Agreed I left that on purpose as I think it's more useful, perhaps we could add the same for files in a future enhancement.

> 2. Re `int` and `null` return values in `directory_filesize()`. Do we need to add unit test for that?

Definitely something that could be unit tested 👍 if you fancy having a bash but not critical to getting this merged I don't think.

dashaluna:
> 1. I can still see `0 of xx MB` for directories that are fully excluded

>Agreed I left that on purpose as I think it's more useful, perhaps we could add the same for files in a future enhancement?

In the last commit we now displaying the `0 of xx KB` for excluded files as well. @willmot do you want to have a look just at the last commit to double check me? And then let's merge this! 🎉

admin/schedule-form-excludes.php
```diff
@@ -300,29 +300,24 @@
...
-												esc_html( $size )
-											);
+										if ( $size ) {
+											$size = size_format( $size );
 										}
 
+										/* translators: 1: Excluded size 2: Overall directory/file size */
+										printf(
+											esc_html__( '%1$s of %2$s', 'backupwordpress' ),
+											esc_html( $excluded_size ),
```
willmot: No need to double escape this

admin/schedule-form-excludes.php
```diff
@@ -300,29 +300,24 @@
...
+											esc_html( $size )
+										);
```
willmot: Or this.

dashaluna:
🎉

Issue #1082, opened 2016-07-21
Update Codecov settings


dashaluna:
As they changed the settings to be in codecov.yml file.

dashaluna:
@willmot I thought I'd start on this so we don't forget. I don't know this things very well, it would be great if you could check if we need:

- [ ] Delete any old settings, to keep things clean.
- [ ] Add any other codecov settings, as I've copied the suggested file from their site.

Thank you!

PR #1083, opened 2016-07-21
New settings file for codecov

dashaluna:
#1082

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  ignore:
+  - tests/.*
+  status:
+    patch:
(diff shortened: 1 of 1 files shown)


dashaluna:
@willmot I thought I'd start on this so we don't forget. I don't know this things very well, it would be great if you could check if we need:

- [ ] Delete any old settings, to keep things clean.
- [ ] Add any other codecov settings, as I've copied the suggested file from their site.

Thank you!

willmot:
The suggested file you copied exactly matches our previous settings so confident this is exactly what we want, thanks!

PR #1090, opened 2016-08-05
Escaping and translator notes to strings

dashaluna:
in `admin` folder


willmot:
@dashaluna just to check if you're still planning on working on this or if it's stale now?

Issue #1089, opened 2016-08-04
Light admin area font in WordPress 4.6


johnbillion:
WordPress 4.6 has changed the font stack used in the admin area. As a result, the light font variant used by BWP is not very legible.

Probably all instances of the `font-weight: lighter;` rule should be removed from the stylesheet.

WordPress 4.5:

<img width="951" alt="screen shot 2016-08-04 at 16 23 12" src="https://cloud.githubusercontent.com/assets/208434/17407649/018638ca-5a60-11e6-95c2-38de19d25680.png">

WordPress 4.6:

<img width="949" alt="screen shot 2016-08-04 at 16 22 22" src="https://cloud.githubusercontent.com/assets/208434/17407655/0a659102-5a60-11e6-9677-31394d5d7bdb.png">

PR #1097, opened 2016-08-26
Stop using lighter font weights as they don't play nice with system fonts

willmot:
Fixes https://github.com/humanmade/backupwordpress/issues/1089

Also removes some unneeded CSS and fixes the styling of the back button on the extension page.

Here's what it now looks like:

screen shot 2016-08-26 at 18 30 16


willmot:
@dashaluna can we change our codecov settings so it shows up as a check rather than posting a comment>

dashaluna:
@willmot I'll look into codecov ^^

Issue #1091, opened 2016-08-12
Add hint about how to change the backup directory to FAQ


COLABORATI:
I can see no obvious way to change the backup directory. On the WordPress plugin site the first item of the FAQ section is about this and it reads: 
`Backups are stored on your server in /wp-content/backups, you can change the directory.`

The expected sentence that should follow this would explain _how to change the directory_ - unfortunately this sentence is missing. 

Please add this sentence to the FAQ after that sentence above: 
`You can change the backup directory by setting the constant HMBKP_PATH to the desired path in your settings file.`

Thanks.

willmot:
I agree it should be easier to change your backup directory!

pdewouters:
I think there's a ticket to add a UI for that

katmoody:
I am adding more documentation to the Knowledge Base for all the constants as well. The information shouldn't be so hard to find, for sure!

katmoody:
This was merged and can be closed. Additional work in #1103 to add in UI section for better handling in the futuer.

PR #1102, opened 2016-08-28
Add some help text for changing your backup path

willmot:
Adds some simple help text to the title tag on the backup path that's shown in the schedule sentence.

screen shot 2016-08-28 at 22 44 26

See https://github.com/humanmade/backupwordpress/issues/1091

### DIFF
--- a/admin/schedule-sentence.php
+++ b/admin/schedule-sentence.php
@@ -73,8 +73,7 @@ switch ( $schedule->get_reoccurrence() ) :
 
 endswitch;
 
-$server = '<span title="' . esc_attr( Path::get_path() ) . '">' . __( 'this server', 'backupwordpress' ) . '</span>';
-$server = '<code>' . esc_attr( str_replace( Path::get_home_path(), '', Path::get_path() ) ) . '</code>';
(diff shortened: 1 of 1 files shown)


dashaluna:
@willmot I'm just wondering if it's clear that the highlighted path could be hovered over for a help tip? Is it possible to add a short text link to open up that help tab? Or perhaps even the word `TIP` or `HELP` with some BG colour and when you click on it, it displays the tip text?

Because it's just a short word, we can add that to UI throughout in the tricky areas. What do you think?

willmot:
> I'm just wondering if it's clear that the highlighted path could be hovered over for a help tip?

Agree it's not very clear, title text is generally not a great way to show help text, but, we already do it in a lot of other places so this really just brings this into line and is a quick, small, improvement.

>  Is it possible to add a short text link to open up that help tab? Or perhaps even the word TIP or `HELP` with some BG colour and when you click on it, it displays the tip text?

Definitely open to exploring this, although have reservations about needing to put help buttons everywhere. I think should be separate to getting this PR in thought.

katmoody:
What about adding a column header with very short description at the top of the column? Rather than a hover-over tip?

PR #1104, opened 2016-08-28
Scrutinizer fixes

willmot:
Fix a bunch of minor issues raised by Scrutinizer.

See https://github.com/humanmade/backupwordpress/issues/1031

### DIFF
--- a/admin/actions.php
+++ b/admin/actions.php
@@ -233,9 +233,7 @@ function edit_schedule_submit() {
 	$schedule = new Scheduled_Backup( sanitize_text_field( $_POST['hmbkp_schedule_id'] ) );
 	$site_size = new Site_Size( $schedule->get_type(), $schedule->get_excludes() );
 
-	$errors = array();
-
-	$settings = array();
+	$errors = $settings = array();
 
 	if ( isset( $_POST['hmbkp_schedule_type'] ) ) {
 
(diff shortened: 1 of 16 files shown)


willmot:
- [ ] Need to fix the new issues added by this PR before merging: https://scrutinizer-ci.com/g/humanmade/backupwordpress/inspections/43b9aa88-fbb4-4a7d-8f67-c92b520c4098/issues/

dashaluna:
@willmot is comment https://github.com/humanmade/backupwordpress/pull/1104#issuecomment-243006112 still stands?

willmot:
Yep

dashaluna:
Cool, I'll wait for more commits to PR :)

willmot:
@sambulance / @shadyvb let's merge this one, it reduces the net number of Scrutinizer issues which is a good thing.

admin/actions.php
```diff
@@ -250,9 +248,9 @@ function edit_schedule_submit() {
 		}
 	}
 
-	if ( isset( $_POST['hmbkp_schedule_recurrence']['hmbkp_type'] ) ) {
+	$schedule_recurrence_type = sanitize_text_field( $_POST['hmbkp_schedule_recurrence']['hmbkp_type'] );
```
shadyvb: I'm not sure why we swapped the lines here, the `if` check should happen before sanitization, so as not to produce an error if the value does not exist.

assets/hmbkp.js
```diff
@@ -139,21 +139,6 @@ var BackUpWordPressAdmin = (function($){
 			);
 		} );
 
-		jQuery( document ).one( 'click', '.hmbkp_send_error_via_email', function ( e ) {
```
shadyvb: This doesn't seem like a CS/CI fix, should probably be double-checked to see if the removed code is not needed / does not need a replacement.

classes/backup/class-backup-status.php
```diff
@@ -18,6 +19,11 @@ public function start( $backup_filename, $status_message ) {
 		$this->set_status( $status_message );
 	}
 
+	/**
+	 * Get the filename of the backup.
+	 *
+	 * @return string The backup filename.
```
shadyvb: `string|null`

assets/hmbkp.js
sambulance: I think this was intentionally removed as unused code, I can't find a reference to either the class `hmbkp_send_error_via_email` or the function `hmbkp_email_error` in the rest of the codebase.

PR #1122, opened 2016-09-28
Extension version text on Extensions page

### LINKED ISSUE #1098: Confusing versions shown when extension updates are available
On the Extensions page, installed extensions are showing a notice that an update is available, even though I'm on the latest version.

screen shot 2016-08-26 at 17 58 40
screen shot 2016-08-26 at 17 58 50

dashaluna:
Fixes #1098 

**TODO**
- [x] Fix Windows Azure icon
- [x] Disable `Update Now` button on Extensions page when no/expired licence. Add note that update needs a valid licence. (**NB** it might be good to add a note to Plugins WP admin page - to notify user and provide a choice to get a licence). **NB**: Added separate issue for this: https://github.com/humanmade/backupwordpress/issues/1130

### DIFF
--- a/admin/extensions.php
+++ b/admin/extensions.php
@@ -2,6 +2,8 @@
 
 namespace HM\BackUpWordPress;
 
+include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+
 ?>
 
 <div class="wrap">
@@ -26,11 +28,14 @@ namespace HM\BackUpWordPress;
 		return strcmp( $b->title->rendered, $a->title->rendered );
 	});
 
-	$installed_plugins = array_reduce( get_plugins(), function( $carry, $item ) {
-		$carry[ strtolower( $item['Name'] ) ] = $item['Version'];
-		return $carry;
-	}, array() );
-
+	$installed_plugins = array();
+	foreach ( get_plugins() as $path => $plugin_info ) {
+		$installed_plugins[ strtolower( $plugin_info['Name'] ) ] = array(
+			'version'   => $plugin_info['Version'],
+			'is_active' => is_plugin_active( $path ),
+			'path'      => $path,
+		);
+	}
 	?>
 
 	<h3><?php esc_html_e( 'Remote Storage', 'backupwordpress' ); ?></h3>
@@ -43,7 +48,11 @@ namespace HM\BackUpWordPress;
 
 			<?php $first = true; ?>
 
-			<?php foreach ( $extensions_data as $extension ) : ?>
+			<?php foreach ( $extensions_data as $extension ) :
+
+				$extension_name_lowcase = strtolower( $extension->title->rendered );
+				$is_extension_installed = in_array( $extension_name_lowcase, array_keys( $installed_plugins ) );
+				?>
 
 				<div class="plugin-card plugin-card-<?php echo esc_attr( $extension->slug ); ?>">
 
@@ -51,7 +60,7 @@ namespace HM\BackUpWordPress;
 
 						<div class="name column-name">
 
-							<h4>
+							<h3>
 
 								<a href="<?php echo esc_url( $extension->link ); ?>" class="thickbox">
 
@@ -61,7 +70,7 @@ namespace HM\BackUpWordPress;
 
 								</a>
 
-							</h4>
+							</h3>
 
 						</div>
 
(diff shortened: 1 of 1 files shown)


willmot:
Thanks @dashaluna, big improvement. Could you drop a screenshot of what it looks like, want to check the new wording feels right in-situ.

dashaluna:
Here is how it looks now:
screen shot 2016-09-28 at 09 50 15

@willmot couple of questions:
1. Is there default WP admin styling class that can be adjusted so that we have more space for plugin version area. So that they are on 2 lines instead of 3? Current styling seems to come from `wp-admin/css` but I couldn't find style guides for it online?

2. I've removed `span with title` attribute here https://github.com/humanmade/backupwordpress/pull/1122/files#diff-44c404adae8631d8d31fdd86aaa82916L151 to simplify the string for translation:
`<span title="<?php echo esc_attr( $extension->modified ); ?>">`
Let me know if that's ok or if I should change it back.

willmot:
The original design intention of this page was to mimic the design of the add plugins screen in wp-admin (seen here: `wp-admin/plugin-install.php?s=query+monitor&tab=search&type=term`).

screen shot 2016-09-28 at 10 06 55

It looks like they no longer even show the version so we could actually just drop that completely from plugins which aren't installed or which don't have an update available.

For extensions which are installed and which need updating, let's replace the installed button with an update button. Something like:

screen shot 2016-09-28 at 10 12 54

We should also change the titles to `<h3 />`'s instead of `<h4 />`'s and remove the CSS `height` from the icons (so they're not stretched).

willmot:
It would obviously be much nicer if we could have much shorter descriptions as well as the cards are all too long and look messy.

dashaluna:
@willmot I've tried to mimic the WP plugin card functionality button as much as possible. So now we have buttons:

1. Update Now <= installed & update available
2. Active <= installed & activated
3. Activate <= installed
4. Buy Now <= otherwise

Couple of questions:

1. WP core differentiate between `Activate` and `Network Activate` ref: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-plugin-install-list-table.php#L469.
2. There are capability checks before displaying plugin management buttons ref https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-plugin-install-list-table.php#L451 What's our approach here? Shall we do the same?

dashaluna:
> and remove the CSS height from the icons (so they're not stretched).

@willmot I can see that rule in the WP native CSS. The icons don't look stretched for me. Are you viewing on big screen? Not sure how to correctly remove this.

willmot:
> 1. WP core differentiate between `Activate` and `Network Activate` ref: 

I'm not actually sure how the extension page functions when BackUpWordPress is Network Activated, something we should test. In Multisite setups the main backup wp-admin page moves to Network Settings so the Extensions page should too and Extensions should then also be Network Activated, let's leave that for a separate ticket though.

2. There are capability checks before displaying... What's our approach here? Shall we do the same?

Good catch, we should do the same 👍

willmot:
> The icons don't look stretched for me.

It's just the Windows Azure icon:
screen shot 2016-09-28 at 14 45 28
Should be:
screen shot 2016-09-28 at 14 45 44

> Not sure how to correctly remove this.

You'll need to override in our CSS file with a more specific selector.

willmot:
Could you also start a ticket for simplifying the descriptions. I think we could probably get them all down to a single paragraph or so.

katmoody:
@willmot and @dashaluna - the descriptions here are actually pulled from each of the product pages - that's why the text changed when I was updating product information. Is there a way to create a custom excerpt to pull for that instead? That would allow granular control over the content displayed, much like the short description pulled for the WP repo example above? 

I'm all for re-writing the product descriptions further as well, but I think we're looking at two different needs here and they each really do need different text - the product page is for people looking for more specific information on each product listing while the description/excerpt is just a short snippet on info on each of them. Thoughts?

willmot:
Agreed best for these to be different text

dashaluna:
@willmot is it a good idea to display sale stats per extension similar to WP ratings? Don't know if it's something we can pull out.

dashaluna:
@willmot I'm a little bit stuck with `Update Now` functionality. I've copied over the update URL from WP core, but I'm guessing because extensions aren't in the WP plugins repo that's why it doesn't work. What should be the process/URL of updating a plugin?

dashaluna:
Update on the `Update Now` functionality :) The update URL works correctly, **only** when the valid licence has been specified prior.

@willmot should we only show `Update Now` on the Extension page only with valid licence, as this is how it currently works for the main WP plugin page ref: https://github.com/humanmade/backupwordpress/issues/1126

willmot:
Yeah good call, we could just add the disabled attribute to the button when the license is invalid.

dashaluna:
@willmot I can't figure out how can I check a license status per extension in the main BWP plugin code. Could you give me a hand with that please?

pdewouters:
I don't think we want to start checking licenses in the main BackUpWordPress plugin

dashaluna:
@pdewouters then there is no way to disable the "Update Now" button if we don't know the license status. And we don't disable it for expired / invalid license when you click the button it says the plugin is up to date, but the button is still displayed. Also, in the case of expired/invalid license we can notify a user and give a choice to purchase it.

dashaluna:
@willmot re `Update Now` button and license checks - I think we should still investigate it but create a new ticket for it, not to block this one. This one already gone further then the original issue description. The functionality works, it's just ever so slightly unclear when you're trying to update an extension that's expired.

I will fix the Azure logo and we should push this to review & merge for the next release.

Also, apologies for a mammoth PR, I know it's over the original description, but seemed to make sense to bundle up related functionality. I tried to make every commit as atomic as possible, so you could review each commit for related functionality. Apologies about that!

willmot:
Awesome Dasha!

>  I think we should still investigate it but create a new ticket for it

Agreed 👍

> I will fix the Azure logo and we should push this to review & merge for the next release.

Great, I think we should do a release next week, we've a few things done now.

> Also, apologies for a mammoth PR, I know it's over the original description

No problem, great to see so much stuff improved!

admin/extensions.php
```diff
@@ -1,165 +1,194 @@
 <?php
-
 namespace HM\BackUpWordPress;
-
+include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
```
willmot: I assume you're including this because we need access to one of the functions it contains? If so, I think it's best practise to include it just before you use the function and add a docblock comment explaining what it's being included for.

That way future developers will know why it's here and whether it's safe to remove or not.

dashaluna: ✅

dashaluna:
@willmot ready for review & merge

willmot:
@dashaluna I'm out today but let's try to release tomorrow.

Issue #1131, opened 2016-10-07
Change disk space calculation restriction to warning


katmoody:
Per #1063 issue - @willmot said: 

> We should definitely account for excludes when calculating the remaining disk space.
> We could also consider making this a warning only instead of blocking actual backups as well. If there isn't enough space then backups will obviously fail anyway.

So we need to change the restriction upon finding the disk space and make it a warning instead of disallowing access to the plugin settings.

sambulance:
If I'm understanding this correctly, we need to remove the [`disk_space_low()` check in the `is_backup_possible()` function](https://github.com/humanmade/backupwordpress/blob/master/functions/core.php#L414). The error message will still appear at the top of the page, but the rest of the functionality will be available.

Created a PR here: https://github.com/humanmade/backupwordpress/pull/1132

libby-barker:
@willmot I saw the you merged the PR related to this. Has it been tested/deployed?

libby-barker:
NM, I see it's in "Ready for..."

willmot:
> tested/deployed?

I guess "deployed" would mean released in a plugin update?

willmot:
Perhaps we should have a ready for release tag.

katmoody:
Is this all released as of 3.6.3.1? @sambulance

sambulance:
This is deployed.

PR #1132, opened 2016-10-20
Remove disk_space_low check from is_backup_possible

sambulance:
Stops low disk space blocking users from accessing BWP settings.

Fixes https://github.com/humanmade/backupwordpress/issues/1131


willmot:
Perfect, thanks @sambulance 👍

Issue #1030, opened 2016-03-12
Allow access to existing backups even if new backup aren't possible


willmot:
Currently if any of the checks in `is_backup_possible` fail we show a warning message and completely disable access to all plugin functionality.

For example if disk space is low you'll see the following:

<img width="1085" alt="screen shot 2016-03-12 at 17 46 27" src="https://cloud.githubusercontent.com/assets/308507/13724362/d0213d5e-e87a-11e5-99b4-b4ef3da9868d.png">

This is more overbearing than we need to be. Instead we should allow access to all functionality except running new backups.
- [ ] Disable running manual backups

willmot:
We should additionally only block access to database and complete backups whilst allowing files only backups if PDO + mysqldump is missing but ZipArchive is available.

katmoody:
Alternatively, if there isn't enough space to do a full site backup, there should also still be the option to back up the database. Reported here: https://wordpress.org/support/topic/your-server-only-has-x-gb-of-disk-space-left-problem?replies=1

SEWilco:
I think this is a bug, not an enhancement issue. Taking away all functionality means users can't adjust settings and delete files so as to fix the problem. (But then, I don't understand why 33% of disk space in my system isn't enough, when that's enough for well over a dozen more backups.)

pdewouters:
also there's a typo : `You're server`

willmot:
> (But then, I don't understand why 33% of disk space in my system isn't enough, when that's enough for well over a dozen more backups.)

@SEWilco that definitely shouldn't be happening. Could you drop a screenshot of the error you're getting?

SEWilco:
@willmot:
Your server only has 9 GB of disk space left which probably isn't enough to complete a backup. Try deleting some existing backups or other files to free up space.

willmot:
@SEWilco and how big is your backup (with and without excludes)?

SEWilco:
@willmot Less than 400MB with excludes. I think. I don't know because I can't make the plugin tell me what my excludes are.

SEWilco:
@willmot Well, that's interesting. I managed to delete enough to make only 47% of disk be in use (15GB available), and now the plugin says: "Backup my database and files (**104 MB**) weekly..." but the backup is **451MB**. The two excluded directories are not in the backup file. DB backup is 30MB (5MB compressed).

SEWilco:
[...]

PR #1133, opened 2016-11-01
#1030 Remove admin block

sambulance:
Remove the check around the admin interface that blocks access if a backup isn't possible. Apply the check instead to the button for triggering a manual backup.

Addresses https://github.com/humanmade/backupwordpress/issues/1030


shadyvb:
@sambulance LGTM, though couple of things to consider:

- Does the user see why the button isn't there ? ie: is there a note somewhere on the same screen ? Should we probably keep a dimmed button but have a small note explaining why it can't be used ?
- I suppose we also do the check within the actual callback ?

sambulance:
@shadyvb 
1. There will be a warning message at the top of the admin screen:
```
Your server only has XX GB of disk space left which probably isn't enough to complete a backup. Try deleting some existing backups or other files to free up space.
```
2. `is_backup_possible()` is checked before a scheduled backup runs

Issue #1063, opened 2016-04-26
Server size is being configured incorrectly, causing backups not to run


katmoody:
Per conversation with @pdewouters - since changing up the function and displaying an error message when the plugin sees the server doesn't have enough space, users are basically getting false positives. Not sure if the function needs rewritten for that or taken out. Might need further information but wanted to list it while we were thinking about it.

willmot:
This should be fixed in https://github.com/humanmade/backupwordpress/pull/1052 which should go out in the 3.6.1 release.

katmoody:
While this seems to fix the issue for most users, we're still getting some folks with false positives. 

from wp.org forum post: https://wordpress.org/support/topic/our-server-only-has-blank-of-disk-space-left?replies=9

katmoody:
From one user: 

> ### WordPress Environment ###
> 
> WP Version: 4.5.2
> WP Multisite: –
> WP Memory Limit: 256 MB
> Language: ru_RU
> 
> ### Server Environment ###
> 
> Server Info: Apache/2.4.7 (Ubuntu)
> PHP Version: 5.5.9-1ubuntu4.11
> PHP Post Max Size: 8 MB
> PHP Time Limit: 0
> PHP Max Input Vars: 1000
> SUHOSIN Installed: –
> MySQL Version: 5.5.44
> Max Upload Size: 8 MB
> Default Timezone is UTC: ✔
> fsockopen/cURL: ✔
> SoapClient: ✔
> DOMDocument: ✔
> GZip: ✔
> Multibyte String: ✔
> 
> BackUpWordPress: от Human Made Limited – 3.6.1

mrhistalk:
I originally had an actual disk space problem caused by having multiple copies of the large weekly backup that seemed to suddenly swell to 135gb per copy. I deleted those files to free up space, but the Tools/Backup panel still wouldn't display anything more than the message that I didn't have enough space free (even though at that point I had plenty of space free -- 485GB at the moment).

This morning I went back into Tools/Backups and it brought up a list of daily backups instead of the error message, so that worked correctly. I then clicked the Weekly tab and again got the "not enough disk space" message. That now happens each time I exit and return to Tools/Backups.

sanya-kuzovlev:
@katmoody I can provide temporary admin access to your website, if it helps to solve the problem.

willmot:
@sanya-kuzovlev that would be really helpful, could you send to tom@humanmade.co.uk. Thanks!

[...]

PR #1135, opened 2016-11-02
#1063 Account for excludes when calculating if disk space is too low

sambulance:
Pass `Excludes` when calculating the site size in `disk_space_low()` function to give a more accurate response.

Previously, `disk_space_low()` would return true if the entire site (including previous backups) was too big to be backed up, not just the files specified to be backed up.

Addresses https://github.com/humanmade/backupwordpress/issues/1063

### DIFF
--- a/functions/interface.php
+++ b/functions/interface.php
@@ -450,7 +450,7 @@ function disk_space_low( $backup_size = false ) {
 
 	if ( ! $backup_size ) {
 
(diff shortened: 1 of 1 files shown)


shadyvb:
@sambulance LGTM, got a tiny comment though, feel free to merge if/when you tackle it.

Issue #1125, opened 2016-09-29
Extensions page: Highlight WP admin Backups menu item when on Extension page


dashaluna:

sambulance:
This wasn't in the pipeline for this sprint, but figured it was a quick one I could pick off easily. Also ensures we get an enhancement in this sprint.

PR: https://github.com/humanmade/backupwordpress/pull/1136

libby-barker:
Awesome @sambulance

shadyvb:
Fixed by #1136

shadyvb:
@sambulance Are we supposed to close this one ?

libby-barker:
Hi @shadyvb it should only be closed if it's deployed.

willmot:
@libby-barker "deployed" in this instance means released to the plugins directory I guess.

willmot:
I guess this can be moved into the 3.6.3 release project?

mikeselander:
Looks like this might have been done already? @katmoody should this still be open?

katmoody:
@mikeselander - looks like it was merged in #1136 -- unsure if that was deployed for sure?

sambulance:
This is done and deployed, can be seen by going to `/wp-admin/tools.php?page=backupwordpress_extensions` and seeing that the Backups menu item in the sidebar is highlighted

PR #1136, opened 2016-11-02
#1125 Highlight the 'Backups' submenu item when on the Extentions page

sambulance:
Forces the main BWP 'Backups' page to be highlighted in the admin sidebar when visiting the 'Extensions' page. 

Achieved by setting the main 'Backups' page to be the current subpage when on the 'Extensions' page via the `parent_file` hook.

Addresses #1125.

### DIFF
--- a/admin/menu.php
+++ b/admin/menu.php
@@ -42,6 +42,28 @@ function extensions() {
 }
 
 /**
+ * Highlights the 'Backups' submenu item when on the Extentions page
+ *
+ * @param string $parent_file The filename of the parent menu item
+ * @return string $parent_file The filename of the parent menu item
+ */
+function highlight_submenu( $parent_file ) {
+
+	global $submenu_file;
+
+	$screen = get_current_screen();
+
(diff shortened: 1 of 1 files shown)


admin/menu.php
```diff
@@ -42,6 +42,29 @@ function extensions() {
...
+	if ( 'tools_page_' . HMBKP_PLUGIN_SLUG . '_extensions' === $screen->id ) {
+
+		// Set the main plugin page to be the active submenu page
+		$submenu_file = HMBKP_PLUGIN_SLUG;
+
+	}
+
+	return $parent_file;
+}
+add_filter( 'parent_file', 'HM\BackUpWordPress\highlight_submenu' );
```
shadyvb: @sambulance Would it be more appropriate to use the `submenu_file` filter instead ? Since you're not using the `parent_file` variable anyway. Would drop the need for the global var as well.

sambulance: Thanks @shadyvb, I did look at using `submenu_file` the first time but couldn't get it to work. Just had another look and got it working this time :) thanks!

PR #1149, opened 2016-12-09
Ensure only new persistant notices are added added to the database

sambulance:
Previously, any new admin notices were merged into the array of persistent notices in the database without checking if the new notice already existed as a persistent notice, causing the database to fill up with multiple instances of the same message. This wasn't seen in the admin because array_unique is used before outputting each notice.


paulgibbs:
LGTM

libby-barker:
Thank you @paulgibbs

PR #1167, opened 2017-03-14
Fix wp-snapshots

katmoody:
This wildcard set is causing wp-snapshots to NOT be excluded per report from user in forum. I confirmed this on my test site, and changed the plugin back to the previous iteration which was just using 'wp-snapshots' - and that worked perfectly. Unsure exactly why this happened but pretty sure the wildcard specification needs further translating for regex to do it correctly?

### DIFF
--- a/classes/class-excludes.php
+++ b/classes/class-excludes.php
@@ -32,7 +32,7 @@ class Excludes {
 		'backupwp',
 		'backwpup-*',
 		'updraft',
-		'(?:wp-)?snapshots?', // wp-snapshots, snapshots, snapshot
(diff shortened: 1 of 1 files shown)


katmoody:
Hey @pdewouters - I tried looking back through the file history for this and I can't figure out at what point this small part of the excludes was changed? But it broke exclusions for the wp-snapshots sub-directories when it was, causing much larger backups. After looking at both the wildcard set on it and the conversion to regex for the excludes which is a bit further down in the same file, I'm pretty sure there is something not quite right there. I'm unsure what that is though?

PR #1172, opened 2017-04-19
Tags and Description update to readme

katmoody:
This is the first pass-through to edit the description to include more
keywords, fix a few awkward sentences, and improve readability. This focuses only on keywords and the description itself for the repo listing.


katmoody:
@noeltock added you as a reviewer here as I wanted to go ahead and push this through since you already had some keyword updates. Can we merge them together?

PR #1175, opened 2017-05-02
1107 remove codecov from plugin

katmoody:
This removes the Codecov.yml file from the plugin to remove the service from integration. I also removed the webhook integration.


katmoody:
@pdewouters - I see by the Codecov notes that I didn't get it all done on my end - what else needs to be done to remove this from the plugin? @willmot thought it should be removed as it's not as helpful. :D Thanks!

roborourke:
Doesn't look to be anything else to do here now. Cheers @katmoody

PR #1177, opened 2017-05-11
Update to error notices. Removed extra extensions upsell and replaced…

katmoody:
… non-dismissible notice with a dismissible notice. Both are at the top, though? @mikeselander you were the one who noticed this previously. My only concern at this point is that both main notices are at the top of the page (the extension one and the rate us one), rather than one at the top and two at the bottom. 

I'm not positive whether having both at the top is desireable. See image: 
screen shot 2017-05-11 at 11 55 19 am

### DIFF
--- a/admin/page.php
+++ b/admin/page.php
@@ -10,7 +10,7 @@ namespace HM\BackUpWordPress;
 
 		BackUpWordPress
 
-		<a class="page-title-action" href="<?php echo esc_url( https://bwp.hmn.md/ ); ?>" target="_blank">Extensions</a>
+        <a class="page-title-action" href="<?php echo esc_url( 'https://bwp.hmn.md/' ); ?>" target="_blank">Extensions</a>
 
 		<?php if ( get_option( 'hmbkp_enable_support' ) ) : ?>
 
@@ -28,9 +28,7 @@ namespace HM\BackUpWordPress;
 	<?php if ( has_server_permissions() ) : ?>
 
 		<?php include_once( HMBKP_PLUGIN_PATH . 'admin/backups.php' ); ?>
-
-		<p class="howto"><?php printf( __( 'If you\'re finding BackUpWordPress useful, please %1$s rate it on the plugin directory%2$s.', 'backupwordpress' ), '<a target="_blank" href="http://wordpress.org/support/view/plugin-reviews/backupwordpress">', '</a>' ); ?></p>
-
+  
 		<?php include_once( HMBKP_PLUGIN_PATH . 'admin/upsell.php' ); ?>
 
 	<?php else : ?>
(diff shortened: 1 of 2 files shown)


katmoody:
Hey @mikeselander thanks so much for the great notes to go along with this - I'll go over them a little later today but really appreciate the time you took to help me out!

admin/page.php
```diff
@@ -10,7 +10,7 @@
 
 		BackUpWordPress
 
-		<a class="page-title-action" href="<?php echo esc_url( https://bwp.hmn.md/ ); ?>" target="_blank">Extensions</a>
+        <a class="page-title-action" href="<?php echo esc_url( 'https://bwp.hmn.md/' ); ?>" target="_blank">Extensions</a>
```
katmoody: Escaping has always been something I get confused easily over, but this makes sense to me. Thanks!

admin/upsell.php
```diff
@@ -1,17 +1,17 @@
...
-<p>
-
 <?php
-/** translators:  the 1st placeholder is the first part of the anchor tag with the link to the extensions admin page and the second is the closing anchor tag */
-printf(
-	__( 'Store your backups securely in the Cloud with %1$sour extensions%2$s', 'backupwordpress' ),
-    '<a href="<?php echo esc_url( https://bwp.hmn.md/ ); ?>" target="_blank">',
-    '</a>'
+/** translators:  the 1st placeholder is the first part of the anchor tag with the link to the plugin review page and the second is the closing anchor tag */
+$cta_message = printf(
```
katmoody: That's the best description I heard on it - I went back and forth a few times and decided someone else would have to be able to tell the difference! Thank you again!

admin/upsell.php
```diff
@@ -1,17 +1,17 @@
...
+	__( 'If you\'re finding BackUpWordPress useful, please %1$s rate it on the plugin directory%2$s.', 'backupwordpress' ),
+	'<a target="_blank" href="http://wordpress.org/support/view/plugin-reviews/backupwordpress">',
+	'</a>'
 );
 ?>
-
-</p>
-
+    <div id="hmbkp-cta-message" class="updated notice is-dismissible">
```
katmoody: Good point. I don't think I adjusted any of the tab/spacing stuff but I likely did without realizing. :(

admin/upsell.php
```diff
@@ -1,17 +1,17 @@
...
+    <p><?php echo wp_kses_post( $cta_message ); ?></p>
+    <button type="button" class="notice-dismiss"><span class="screen-reader-text"><?php esc_html_e( 'Dismiss this notice.', 'backupwordpress' ); ?></span></button>
 </div>
+
```
katmoody: *hides head in shame* Oops.

katmoody: Actually, @mikeselander - that second div closes the one at the very top. Should it simple be fully removed? There is CSS attached to it, I think.

katmoody:
Yay! Thanks so much @mikeselander !!!!

Issue #1045, opened 2016-03-28
Error when Uninstalling after Nonsupported php version


katmoody:
So a user wasn't able to use the plugin after I getting a message saying they needed php 5.3.2 or later and would not install, and they went to delete it, getting this error: 

`Parse error: syntax error, unexpected T_STRING in /home/lollipop/public_html/wp-content/plugins/backupwordpress/uninstall.php on line 16
`

Not positive if it has been addressed or not but wanted to report it just in case. :D

willmot:
Ah, we should make `uninstall.php` PHP 5.2 compatible.

pdewouters:
I think that's an old version though

willmot:
@pdewouters what makes you say that?

pdewouters:
line 16 is a blank line in the current version

roborourke:
Used https://github.com/wimg/PHPCompatibility to make sure there was nothing else.

PR #1184, opened 2017-09-07
5.2 compat for uninstall

roborourke:
pulled out bits from the class method

Related #1045


roborourke:
Closing PR out as seems stale now.

PR #1186, opened 2018-02-19
Deploy process new

pdewouters:
Eliminate generation of readme.txt FAQ from faq.php - adds complexity with few benefits, as this is rarely/never updated. It doesn’t take long to update in both places.

Remove all grunt tasks related to deploying to WP.org, as we will use https://github.com/GaryJones/wordpress-plugin-svn-deploy instead.

Remove JSHint linting, deprecated

Change assets folder to CSS and JS separate folders. As “assets” clashes with the deploy script.
Updated references in paths

Remove the grunt subfolder as the task that used that is unmaintained and ads unnecessary complexity.

Removed any references to bower as it’s unmaintained ( prefer yarn )

Removed release script

Removed autoprefixer, unmaintained


pdewouters:
The deploy bash script needs to be installed one level up from the plugin folder, so under `plugins/`

PR #1189, opened 2018-03-05
Update package-lock.json

joemcgill:
This removes a bunch of outdated dependencies from the lock file.


kadamwhite:
@joemcgill Why not remove the lockfile wholesale?

pdewouters:
@kadamwhite I thought the lock file should be committed?

katmoody:
Thank you guys for working on this - is this due to the WP.org forum post? @joemcgill @pdewouters or @kadamwhite

pdewouters:
@katmoody we get the notice from Github as it's an automatic scan

katmoody:
So I have one more stupid question  @joemcgill @pdewouters or @kadamwhite --
Is this updated version okay to release in WP.org? Or would I still need to update version numbers to do that?

joemcgill:
@katmoody the issue in that forum post looks to be the same one we've addressed here. Technically, you should probably bump the version number to 3.6.5 before pushing this to the .org repo, but I don't see it being very time sensitive as this won't affect most people who are installing directly from that repo, so you could wait an package that up with other bug fixes if there are some in progress.

katmoody:
I think we can potentially have something to push in the next week or so if I take a look at outstanding issues a bit. Thanks @joemcgill!

